PR #11, opened 2020-02-04
Use FDM call instead of jvp in rrule_test

mattBrzezinski:


oxinabox:
can we add a test?

mattBrzezinski:
> can we add a test?

Do you have any suggestions on tests to add? I tried pulling some from `ChainRules.jl` but it's back to the `iterate(::Nothing)` issue.

oxinabox:
I do not.

I guess can merge without.
Since it is restoring to what it was originally

But we need to build up to good test coverage before we can make enhancements

PR #13, opened 2020-02-05
Check that rrule x_ad DNE

mattBrzezinski:
Reverting this completely back to the original from `ChainRules/test/test_utils.jl`


src/ChainRulesTestUtils.jl
```diff
@@ -175,8 +175,13 @@ function rrule_test(f, ȳ, xx̄s::Tuple{Any, Any}...; rtol=1e-9, atol=1e-9, fdm
 
     # Correctness testing via finite differencing.
     x̄s_fd = _make_fdm_call(fdm, f, ȳ, xs, x̄s .== nothing)
-    map(x̄s_ad, x̄s_fd) do x̄_ad, x̄_fd
-        @test isapprox(x̄_ad, x̄_fd; rtol=rtol, atol=atol, kwargs...)
+    for (x̄_ad, x̄_fd) in zip(x̄s_ad, x̄s_fd)
+        if x̄_fd === nothing
+            # The way we've structured the above, this tests that the rule is a DoesNotExistRule
```
oxinabox:
```suggestion
            # The way we've structured the above, this tests that the propagator is returning a  DoesNotExist
```

PR #21, opened 2020-03-01
Add docstring for `_make_fdm_call`

nickrobinson251:
- Explain that this is basically just `FiniteDifferences.j'vp`
- Doesn't go as far as to justify having it, or why it's written so sneakily 😂

### DIFF
--- a/src/ChainRulesTestUtils.jl
+++ b/src/ChainRulesTestUtils.jl
@@ -14,6 +14,22 @@ export test_scalar, frule_test, rrule_test, isapprox, generate_well_conditioned_
 Base.isapprox(d_ad::DoesNotExist, d_fd; kwargs...) = error("Tried to differentiate w.r.t. a `DoesNotExist`")
 Base.isapprox(d_ad::AbstractDifferential, d_fd; kwargs...) = isapprox(extern(d_ad), d_fd; kwargs...)
 
+"""
+    _make_fdm_call(fdm, f, ȳ, xs, ignores) -> Tuple
+
+Call `FiniteDifferences.j′vp`, with the option to ignore certain `xs`.
+
+# Arguments
+- `fdm::FiniteDifferenceMethod`: How to numerically differentiate `f`.
+- `f`: The function to differentiate.
+- `ȳ`: adjoint w.r.t. output of `f`.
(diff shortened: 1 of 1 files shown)


src/ChainRulesTestUtils.jl
```diff
@@ -14,6 +14,22 @@ export test_scalar, frule_test, rrule_test, isapprox, generate_well_conditioned_
...
+# Arguments
+- `fdm::FiniteDifferenceMethod`: How to numerically differentiate `f`.
+- `f`: The function to differentiate.
+- `ȳ`: adjoint w.r.t. output of `f`.
+- `xs`: Inputs to `f`, such that `y = f(xs...)`.
+- `ignores`: Collection of `Bool`s, the same length as `xs`.
+  If `ignores[i] === true`, then `xs[i]` is ignored; `∂f_∂xs[i] === nothing`.
+
+# Returns
+- `∂f_∂xs::Tuple`: Derivatives estimated by finite differencing.
```
willtebbutt: `∂f_∂xs` should be `dxs`, since it's not computing the gradient w.r.t. the output.

src/ChainRulesTestUtils.jl
```diff
@@ -14,6 +14,22 @@ export test_scalar, frule_test, rrule_test, isapprox, generate_well_conditioned_
 Base.isapprox(d_ad::DoesNotExist, d_fd; kwargs...) = error("Tried to differentiate w.r.t. a `DoesNotExist`")
 Base.isapprox(d_ad::AbstractDifferential, d_fd; kwargs...) = isapprox(extern(d_ad), d_fd; kwargs...)
 
+"""
+    _make_fdm_call(fdm, f, ȳ, xs, ignores) -> Tuple
+
+Call `FiniteDifferences.j′vp`, with the option to ignore certain `xs`.
+
+# Arguments
+- `fdm::FiniteDifferenceMethod`: How to numerically differentiate `f`.
```
mattBrzezinski: We specify the type of `fdm` but in the actual function signature we do not.

nickrobinson251: It's true that for this function to the work we need it to be this type

But it's not type constrained in the definition because we don't dispatch on it

This docstring is anyway just here to help future contributors who will most likely be replacing this function

PR #26, opened 2020-04-28
Reorganize all the files

oxinabox:
This PR makes no changes except:
 - moving code
 - adding outer testsets
 - rename `ensure_running_in_functor` to `_ensure_running_in_functor`

Out of scope for code review on this is:
 - Content of the code or tests (since they were not changed)

Inscope for code review is:
 - the new organization of the code


src/data_generation.jl
```diff
@@ -0,0 +1,6 @@
+# Useful for LinearAlgebra tests
```
oxinabox, line 1: This got its own file because I forsee that we will need a lot more of these to generate various kinds of structured matrixes, composites etc.

src/testers.jl
```diff
@@ -0,0 +1,180 @@
+function _ensure_not_running_on_functor(f, name)
```
oxinabox, line 1: I am tempted to break this up futher but I am not sure how.
Maybe
`/testers/common.jl`, `/testers/frule.jl`, `/testers/rrule.jl`, `/testers/scalarrule.jl` 
?

nickrobinson251: this seems good for now

I think being able to see all these in one place might be helpful for refactor work that's likely to happen

PR #27, opened 2020-04-29
Generate well conditioned matrix using Random.GLOBAL_RNG

mattBrzezinski:
<s>We need to remove the random number generation parameter in the generate_well_conditioned_matrix` function to be able to continue with https://github.com/JuliaDiff/ChainRules.jl/issues/180</s>

Create a secondary definition for `generate_well_conditioned_matrix()`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.2.2"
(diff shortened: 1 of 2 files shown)


mattBrzezinski:
Closing this since there is a neat `Random.GLOBAL_RNG` which we can use!

nickrobinson251:
we could still add the method `generate_well_conditioned_matrix(N) = generate_well_conditioned_matrix(Random.GLOBAL_RNG, N)`

PR #28, opened 2020-04-29
UndefVarError Random not defined in data_generation.jl

mattBrzezinski:
```
 UndefVarError: Random not defined
  Stacktrace:
   [1] generate_well_conditioned_matrix(::Int64) at /Users/mattbr/.julia/packages/ChainRulesTestUtils/pVYVF/src/data_generation.jl:7
```

We don't have any tests for `data_generation.jl`, in `0.2.3` I added in a secondary function for `generate_well_conditioned_matrix(N)` which used the `GLOBAL_RNG` from `Random`. It was not defined in this source file and would error out.

I added in some tests for a pre-defined rng and the global one and made the changes so that this code would work.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,9 +15,3 @@ ChainRulesCore = "0.7.1"
 Compat = "3"
 FiniteDifferences = "0.9"
 julia = "1"
-
-[extras]
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
-
-[targets]
-test = ["Test"]
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.2.3"
+version = "0.2.4"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -7,6 +7,7 @@ ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
@@ -16,8 +17,7 @@ FiniteDifferences = "0.9"
 julia = "1"
 
 [extras]
-Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Random", "Test"]
+test = ["Test"]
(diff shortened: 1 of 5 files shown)


src/data_generation.jl
```diff
@@ -1,3 +1,5 @@
+using Random
```
nickrobinson251, line 1: let's keep all imports in `src/ChainRulesTestUtils`

test/data_generation.jl
```diff
@@ -0,0 +1,14 @@
+@testset "Data Generation" begin
+    @testset "Generate Well Conditioned Matrix - RNG" begin
+        rng = MersenneTwister(1)
+        matrix = generate_well_conditioned_matrix(rng, 5)
+
+        @test isempty(matrix) == false
```
nickrobinson251, line 6: let's test the "well conditioned" property (here and below too)... which should probably involve some test using `cond`, but i think testing that the matrix is positive definite is probably sufficient

```suggestion
        @test !isempty(matrix)
        @test isposdef(matrix)
```

test/data_generation.jl
```diff
@@ -0,0 +1,14 @@
+@testset "Data Generation" begin
+    @testset "Generate Well Conditioned Matrix - RNG" begin
```
nickrobinson251, line 2: alternatively, you could nest testsets again
```julia
@testset "Generate Well Conditioned Matrix" begin
    @testset "pass in RNG" begin
         ...
    @testset "don't pass RNG" begin
```

src/data_generation.jl
nickrobinson251: Also, `Random` needs to be moved to the `[deps]` in Project.toml (and removed from `[extras]` and `[targets]`)

Project.toml
```diff
@@ -16,8 +17,7 @@ FiniteDifferences = "0.9"
 julia = "1"
 
 [extras]
-Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Random", "Test"]
+test = ["Test"]
```
nickrobinson251, line 23: just noticed in passing - unrelated to this PR, but maybe nice to change - since `Test` is already in `deps` we don't need it here, and can remove `targets/extras` sections completely.

oxinabox:
We can actually test if it is well conditioned.
```julia
using LinearAlgebra

@test cond(matrix) < 20  # for ill conditoned 5x5 matrix will be >> 20
...

I think this is right

**Comment by nickrobinson251** [comment]
> I think this is right

I didn't know what number was right here... where'd you get 20 from?

**Comment by mattBrzezinski** [comment]
I'll keep this open for now while we discuss additional testing.

**Comment by oxinabox** [comment]
The 20 comes from a bad place. I generated 500,000 well conditioned matrixes using this method and the highest condition number for any of them was about 16.

I generated a bunch of other matrixes via various similar products and the Condition number was between 1,000 and 1e9.
Wikipedia just says for a well conditioned matrix the condition number is "small"

File: test/data_generation.jl [hunk 5]
```diff
@@ -1,18 +1,23 @@
 @testset "Data Generation" begin
+    function _is_well_conditioned(matrix::Array{Float64})
+        @test !isempty(matrix)
+        @test isposdef(matrix)
+        @test cond(matrix) < 20
```
nickrobinson251, line 5: else the `20` number is truly magic 
```suggestion
        @asset length(matrix) ≤ 25  
        @test cond(matrix) < 20
```

test/data_generation.jl
```diff
@@ -0,0 +1,23 @@
+@testset "Data Generation" begin
+    function _is_well_conditioned(matrix::Array{Float64})
```
nickrobinson251, line 2:
```suggestion
    function _is_well_conditioned(matrix)
```

nickrobinson251:
this is a big improvement, and we can always improve the `cond` test in future. thanks a lot! merge when you're ready

PR #43, opened 2020-06-25
Automatic tangent generation for simple types

willtebbutt:
First steps of an implementation to automatically construct appropriate tangent vectors. @oxinabox @sethaxen let me know if you think these are okay as a starting point / if you think that there are any other types that are crucial to have this implemented for.

I would very much like to cover general structs with a generated function, but I've not had a chance to sort that out yet.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.3"
+version = "0.3.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
@@ -13,5 +13,5 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 [compat]
 ChainRulesCore = "0.8"
 Compat = "3"
-FiniteDifferences = "0.9"
+FiniteDifferences = "0.9, 0.10"
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.4.0"
+version = "0.4.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
-ChainRulesCore = "0.8"
+ChainRulesCore = "0.9.1"
 Compat = "3"
 FiniteDifferences = "0.9, 0.10"
 julia = "1"
--- a/src/generate_tangent.jl
+++ b/src/generate_tangent.jl
@@ -3,7 +3,7 @@
 
 Returns a randomly generated tangent vector appropriate for the primal value `x`.
 """
-rand_tangent(x) = rand_tangent(Random.default_rng(), x)
+rand_tangent(x) = rand_tangent(Random.GLOBAL_RNG, x)
 
 function rand_tangent(rng::AbstractRNG, x::Union{Symbol, AbstractChar, AbstractString})
     return DoesNotExist()
--- a/src/generate_tangent.jl
+++ b/src/generate_tangent.jl
@@ -1,0 +1,13 @@
+rand_tangent(rng::AbstractRNG, x::Integer) = Zero()
+
+rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = randn(rng, T)
+
+rand_tangent(rng::AbstractRNG, x::StridedArray) = rand_tangent.(rng, x)
+
+function rand_tangent(rng::AbstractRNG, x::T) where {T<:Tuple}
+    return Composite{T}(rand_tangent.(rng, x)...)
+end
+
+function rand_tangent(rng::AbstractRNG, xs::T) where {T<:NamedTuple}
+    return Composite{T}(;map(x -> rand_tangent(rng, x), xs)...)
+end
--- a/src/generate_tangent.jl
+++ b/src/generate_tangent.jl
@@ -13,7 +13,7 @@ rand_tangent(rng::AbstractRNG, x::Integer) = DoesNotExist()
 
 rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = randn(rng, T)
 
-rand_tangent(rng::AbstractRNG, x::StridedArray) = rand_tangent.(rng, x)
+rand_tangent(rng::AbstractRNG, x::StridedArray) = rand_tangent.(Ref(rng), x)
 
 function rand_tangent(rng::AbstractRNG, x::T) where {T<:Tuple}
     return Composite{T}(rand_tangent.(Ref(rng), x)...)
--- a/src/generate_tangent.jl
+++ b/src/generate_tangent.jl
@@ -1,4 +1,10 @@
-rand_tangent(rng::AbstractRNG, x::Integer) = Zero()
+rand_tangent(x) = rand_tangent(Random.GLOBAL_RNG, x)
+
+function rand_tangent(rng::AbstractRNG, x::Union{Symbol, AbstractChar, AbstractString})
+    return DoesNotExist()
+end
+
+rand_tangent(rng::AbstractRNG, x::Integer) = DoesNotExist()
 
 rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = randn(rng, T)
 
@@ -9,5 +15,5 @@ function rand_tangent(rng::AbstractRNG, x::T) where {T<:Tuple}
 end
 
 function rand_tangent(rng::AbstractRNG, xs::T) where {T<:NamedTuple}
-    return Composite{T}(;map(x -> rand_tangent(rng, x), xs)...)
+    return Composite{T}(; map(x -> rand_tangent(rng, x), xs)...)
 end
--- a/src/generate_tangent.jl
+++ b/src/generate_tangent.jl
@@ -24,6 +24,7 @@ function rand_tangent(rng::AbstractRNG, xs::T) where {T<:NamedTuple}
 end
 
 function rand_tangent(rng::AbstractRNG, x::T) where {T}
+    isstructtype(T) || ArgumentError("Non-struct types are not supported by this fallback.")
     field_names = fieldnames(T)
     if length(field_names) > 0
         tangents = map(field_names) do field_name
--- a/src/generate_tangent.jl
+++ b/src/generate_tangent.jl
@@ -24,7 +24,10 @@ function rand_tangent(rng::AbstractRNG, xs::T) where {T<:NamedTuple}
 end
 
 function rand_tangent(rng::AbstractRNG, x::T) where {T}
-    isstructtype(T) || ArgumentError("Non-struct types are not supported by this fallback.")
+    if !isstructtype(T)
+        throw(ArgumentError("Non-struct types are not supported by this fallback."))
+    end
+
     field_names = fieldnames(T)
     if length(field_names) > 0
         tangents = map(field_names) do field_name
--- a/src/generate_tangent.jl
+++ b/src/generate_tangent.jl
@@ -1,3 +1,8 @@
+"""
+    rand_tangent([rng::AbstractRNG], x)
+
+Returns a randomly generated tangent vector appropriate for the primal value `x`.
+"""
 rand_tangent(x) = rand_tangent(Random.GLOBAL_RNG, x)
 
 function rand_tangent(rng::AbstractRNG, x::Union{Symbol, AbstractChar, AbstractString})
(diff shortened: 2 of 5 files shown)


oxinabox:
> I would very much like to cover general structs with a generated function, but I've not had a chance to sort that out yet.

I donb't think it should need a generated function.
It should optimize just fine with a normal function using `fieldnames` etc.
And even if it doesn't it is just not that performance critical

src/generate_tangent.jl
```diff
@@ -0,0 +1,13 @@
+rand_tangent(rng::AbstractRNG, x::Integer) = Zero()
```
oxinabox, line 1: This is a hard call.

src/generate_tangent.jl
```diff
@@ -0,0 +1,13 @@
...
+
+rand_tangent(rng::AbstractRNG, x::StridedArray) = rand_tangent.(rng, x)
+
+function rand_tangent(rng::AbstractRNG, x::T) where {T<:Tuple}
+    return Composite{T}(rand_tangent.(rng, x)...)
+end
+
+function rand_tangent(rng::AbstractRNG, xs::T) where {T<:NamedTuple}
+    return Composite{T}(;map(x -> rand_tangent(rng, x), xs)...)
+end
```
oxinabox, line 13: We should add default to not needing a RNG

```suggestion
end

rand_tangent(x) = rand_tangent(Random.default_rng(), x)
```

src/generate_tangent.jl
```diff
@@ -0,0 +1,13 @@
+rand_tangent(rng::AbstractRNG, x::Integer) = Zero()
+
```
oxinabox, line 2: How about for things that are not pertubable

```suggestion

rand_tangent(rng::AbstractRNG, x::Union{Symbol, AbstractChar, AbstractString} = DoesNotExist()

```

src/generate_tangent.jl
```diff
@@ -0,0 +1,13 @@
...
+rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = randn(rng, T)
```
oxinabox, line 3: I wonder if this makes it too zero centered?

Should we make if offzero?
```suggestion
rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = 100randn(rng, T) + 10rand(rng, T)
```

oxinabox:
I think this needs a docstring that describes `x` as an instance of the primal value

test/generate_tangent.jl
```diff
@@ -0,0 +1,30 @@
+using ChainRulesTestUtils: rand_tangent
+
+@testset "generate_tangent" begin
```
oxinabox, line 3: We should test that 
```
rand_tangent(x) + x isa typeof(x)
```

src/generate_tangent.jl
willtebbutt: Haha yeah. Not sure that I've got this right tbh. Depends whether we want to treat the integers as being embedded in the real line, or as being their own thing 🤷

src/generate_tangent.jl
willtebbutt: Good shout. Will push a change.

src/generate_tangent.jl
willtebbutt: I've never found this to be a problem. Generally speaking low-variance values aren't a bad bet because large values tend to stress out numerics. I mean, obviously you can produce bad numerics with what I've proposed, but I would expect high variance stuff to produce more problems than it solves tbh.

src/generate_tangent.jl
sethaxen: I'm in favor of treating integers as integers. Unless I'm missing something, the only reason to treat integers as embedded in the reals is so that if someone doesn't bother to represent an input to a rule or to a function like `Zygote.gradient` with a float, it still gives them a reasonable answer. Otherwise, integers will be introduced as hard-coded values whose adjoints are unused or quantities with zero derivatives (e.g. produced by `round`).

And then there are cases like the adjoint of `p::Integer` in `y=x^p`. If you treat `p` as embedded in the reals, then the pullback must produce a complex adjoint on `p`, since `x` could be negative, even though for integer `p`, the function takes reals and produces a real. But if you treat `p` as an integer, then it has an adjoint of `Zero()`. To me, the latter seems more like the right thing than the former.

src/generate_tangent.jl
sethaxen: Perhaps `Bool` too?

src/generate_tangent.jl
willtebbutt: I've doubled down on this now, going with `DoesNotExist`.

willtebbutt:
@oxinabox I've got a weird test failure. Would you mind taking a look? It's in some of your more complex code, so might be obvious to you what's going on.

src/generate_tangent.jl
oxinabox: Ok fair enough

oxinabox:
Which error?

https://travis-ci.com/github/JuliaDiff/ChainRulesTestUtils.jl/jobs/354027267#L263-L264
`Int + DoesNotExist()` erroring seems wrong.
Answer should be an `Int`
code for that is here: https://github.com/JuliaDiff/ChainRulesCore.jl/blob/master/src/differential_arithmetic.jl#L20

---

Errors on 1.3 and earlier about `length` are because you are broadcasting a `RNG` which is not supported until 1.4
https://github.com/JuliaLang/julia/pull/33213

Could port that change into Compat.jl
or could just add `randn.(Ref(rng), x)`

willtebbutt:
Nah, this error:
```julia
generate_tangent: Error During Test at /Users/willtebbutt/.julia/dev/ChainRulesTestUtils/test/generate_tangent.jl:25
  Test threw exception
  Expression: x + rand_tangent(rng, x) isa typeof(x)
  MethodError: Cannot `convert` an object of type Int64 to an object of type DoesNotExist
  Closest candidates are:
    convert(::Type{T}, !Matched::T) where T at essentials.jl:171
  Stacktrace:
   [1] convert(::Type{Tuple{DoesNotExist}}, ::Tuple{Int64}) at ./essentials.jl:310 (repeats 2 times)
   [2] Tuple{Float64,DoesNotExist}(::Tuple{Float64,Int64}) at ./tuple.jl:225
   [3] NamedTuple{(:a, :b),Tuple{Float64,DoesNotExist}}(::Tuple{Float64,Int64}) at ./namedtuple.jl:72
   [4] macro expansion at /Users/willtebbutt/.julia/packages/ChainRulesCore/Q5Nrj/src/differentials/composite.jl:0 [inlined]
   [5] elementwise_add(::NamedTuple{(:a, :b),Tuple{Float64,Int64}}, ::NamedTuple{(:a, :b),Tuple{Float64,DoesNotExist}}) at /Users/willtebbutt/.julia/packages/ChainRulesCore/Q5Nrj/src/differentials/composite.jl:191
   [6] +(::NamedTuple{(:a, :b),Tuple{Float64,Int64}}, ::Composite{NamedTuple{(:a, :b),Tuple{Float64,Int64}},NamedTuple{(:a, :b),Tuple{Float64,DoesNotExist}}}) at /Users/willtebbutt/.julia/packages/ChainRulesCore/Q5Nrj/src/differential_arithmetic.jl:88
   [7] (::var"#3#4"{MersenneTwister})(::Tuple{NamedTuple{(:a, :b),Tuple{Float64,Int64}},UnionAll}) at /Users/willtebbutt/.julia/dev/ChainRulesTestUtils/test/generate_tangent.jl:25
   [8] foreach(::var"#3#4"{MersenneTwister}, ::Array{Tuple{Any,Type},1}) at ./abstractarray.jl:1919
   [9] top-level scope at /Users/willtebbutt/.julia/dev/ChainRulesTestUtils/test/generate_tangent.jl:6
   [10] top-level scope at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Test/src/Test.jl:1113
   [11] top-level scope at /Users/willtebbutt/.julia/dev/ChainRulesTestUtils/test/generate_tangent.jl:4
```
Appears to stem from the `Composite` code.

oxinabox:
AH yes that is a bug in ChainRulesCore. I will fix it

Project.toml
```diff
@@ -11,7 +11,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
-ChainRulesCore = "0.8"
+ChainRulesCore = "0.9.1"
```
sethaxen, line 14: Does this bump mean this PR is blocked by #44?

src/generate_tangent.jl
```diff
@@ -0,0 +1,24 @@
+"""
+    rand_tangent([rng::AbstractRNG], x)
```
sethaxen, line 2:
```suggestion
    rand_tangent([rng::AbstractRNG,] x)
```

willtebbutt:
@oxinabox your fix in ChainRulesCore appears to have fixed the issue. Thanks for the quick fix.

src/generate_tangent.jl
```diff
@@ -0,0 +1,24 @@
...
+end
+
+rand_tangent(rng::AbstractRNG, x::Integer) = DoesNotExist()
+
+rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = randn(rng, T)
+
+rand_tangent(rng::AbstractRNG, x::StridedArray) = rand_tangent.(rng, x)
+
+function rand_tangent(rng::AbstractRNG, x::T) where {T<:Tuple}
+    return Composite{T}(rand_tangent.(rng, x)...)
```
sethaxen, line 19: From @oxinabox's suggestion
```suggestion
    return Composite{T}(rand_tangent.(Ref(rng), x)...)
```

Project.toml
willtebbutt: Hmmm possibly. There's no particular hurry with this though, so I'm happy to wait until it's merged.

sethaxen:
Unless there's more you want to add, I think all this is missing is `rand_tangent(::AbstractRNG, ::Bool)`.

willtebbutt:
> Unless there's more you want to add, I think all this is missing is rand_tangent(::AbstractRNG, ::Bool).

I think `Bool` should be covered by `Integer`. I'll add a test though.

willtebbutt:
I also still definitely want to add stuff that can handle `Function`s, and more general `struct`s. That should happen today at some point.

sethaxen:
> > Unless there's more you want to add, I think all this is missing is rand_tangent(::AbstractRNG, ::Bool).
> 
> I think `Bool` should be covered by `Integer`. I'll add a test though.

TIL that `Bool <: Integer`

willtebbutt:
@sethaxen @oxinabox I'm happy with this PR now. Let me know if there's anything else you want.

src/generate_tangent.jl
```diff
@@ -0,0 +1,36 @@
+"""
+    rand_tangent([rng::AbstractRNG,] x)
+
+Returns a randomly generated tangent vector appropriate for the primal value `x`.
+"""
+rand_tangent(x) = rand_tangent(Random.GLOBAL_RNG, x)
```
oxinabox, line 6: Base internally uses:
```suggestion
rand_tangent(x) = rand_tangent(Random.default_rng(), x)
```

src/generate_tangent.jl
```diff
@@ -0,0 +1,36 @@
...
+
+function rand_tangent(rng::AbstractRNG, x::T) where {T<:Tuple}
+    return Composite{T}(rand_tangent.(Ref(rng), x)...)
+end
+
+function rand_tangent(rng::AbstractRNG, xs::T) where {T<:NamedTuple}
+    return Composite{T}(; map(x -> rand_tangent(rng, x), xs)...)
+end
+
+function rand_tangent(rng::AbstractRNG, x::T) where {T}
```
oxinabox, line 26: Probably we only should be supporting struct types.
Not `primative` types.
```suggestion
function rand_tangent(rng::AbstractRNG, x::T) where {T}
    isstructtype(T) || ArgumentError("Non-struct types are not supported by this fallback.")
```

test/generate_tangent.jl
```diff
@@ -0,0 +1,37 @@
...
+        (randn(Complex{Float64}, 2), Vector{Complex{Float64}}),
+        (randn(5, 4), Matrix{Float64}),
+        (randn(Complex{Float32}, 5, 4), Matrix{Complex{Float32}}),
+        ([randn(5, 4), 4.0], Vector{Any}),
+        ((4.0, ), Composite{Tuple{Float64}}),
+        ((5.0, randn(3)), Composite{Tuple{Float64, Vector{Float64}}}),
+        ((a=4.0, ), Composite{NamedTuple{(:a,), Tuple{Float64}}}),
+        ((a=5.0, b=1), Composite{NamedTuple{(:a, :b), Tuple{Float64, Int}}}),
+        (sin, typeof(NO_FIELDS)),
+        (Foo(5.0, 4, rand(rng, 3)), Composite{Foo}),
```
sethaxen, line 31: Nested structs are supported, right? Could you test something like `Foo(5.0, 4, Foo(2.0, 3, 'a'))`?

willtebbutt:
Thanks for the comments, they're now incorporated. Ready for another round of reviews @sethaxen @oxinabox

willtebbutt:
@oxinabox `default_rng` wasn't a thing on 1.0, and `Compat` doesn't appear to have support for it. Do you have any suggestions, or shall I revert to `GLOBAL_RNG`?

oxinabox:
Revert to `GLOBAL_RNG` i think.

Issue #56, opened 2020-10-02
Support for testing InplaceThunks


oxinabox:
We used to have something like this, but when we removed accumulate etc we removed it.

To do this well it would be good to bring back `accumulate!`
See https://github.com/JuliaDiff/ChainRulesCore.jl/issues/113

oxinabox:
closed by https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/59

PR #57, opened 2020-10-05
Overhall equality checks, to give better failure messages

### LINKED ISSUE #7: Remove automatic `extern` in `isapprox`?
This may be the same issue as "do we want `extern` at all?" (https://github.com/JuliaDiff/ChainRulesCore.jl/issues/56)

But I think if we have it, we may need it to be "smarter" if we're going to have `isapprox(a::Differential, b) = isapprox(extern(a), b)` as we currently do.

Because we will need to `extern(a)` (or maybe just `unthunk`?) the differential type _before_ calling `isapprox` anyway if `a` it is not a scalar (e.g. the `Zero` Matix case in SVD tests: https://github.com/JuliaDiff/ChainRules.jl/blob/8c4985a9a941df165740badfbee83cdf93661742/test/rulesets/LinearAlgebra/factorization.jl#L20-L22)

oxinabox:
<s>I wanted to add accumulation checks,
but to do that  </s>
I felt it was good to refactor away from using `isapprox` as much as we do,
<s>I am not entirely sure this PR is a great idea,
the accumulation stuff is easy enough to put in another PR, if we don't like the rest of it.
</s>
This PR is a great idea.
To deal with giving useful error messages an not things about thunks not being equal.
and getting rid of all the hacks with `extern`, `collect`, and `isapprox`.
effectively closes #7  as we will be able to delete `isapprox` on differentials with the next major release


src/iterator.jl
```diff
@@ -43,6 +43,14 @@ function Base.hash(iter::TestIterator{<:Any,IT,IS}) where {IT,IS}
     return mapreduce(hash, hash, (iter.data, IT, IS))
 end
 
+# To make it a valid differential: needs at very least `zero` and `+`
+Base.zero(::Type{<:TestIterator}) = Zero()
+function Base.:+(iter1::TestIterator{T,IS,IE}, iter2::TestIterator{T,IS,IE}) where {T,IS,IE}
+    return TestIterator{T,IS,IE}(map(+, iter1, iter2))
+end
```
oxinabox, line 50: We were using this like a differential, without these being defined,
which ment that testing accumulation errored.

src/check_result.jl
```diff
@@ -0,0 +1,51 @@
+# For once you have the sensitivity by two methods (e.g  both finite-differencing and  AD)
```
oxinabox, line 1: Basically this file replaces `isapprox.jl`
though the removal of `isapprox.jl` is not done yet

src/testers.jl
```diff
@@ -126,9 +126,9 @@ function test_scalar(f, z; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(),
         frule_test(f, (z, Δx); rtol=rtol, atol=atol, fdm=fdm, fkwargs=fkwargs, kwargs...)
         if z isa Complex
             # check that same tangent is produced for tangent 1.0 and 1.0 + 0.0im
-            @test isapprox(
-                frule((Zero(), real(Δx)), f, z; fkwargs...)[2],
-                frule((Zero(), Δx), f, z; fkwargs...)[2],
+            check_equal(
+                frule((Zero(), real(Δx)), f, z; fkwargs...)[2]::Number,
+                frule((Zero(), Δx), f, z; fkwargs...)[2]::Number,
```
oxinabox, line 131: scalar rules shout return `Numbers`.
but maybe this should be moved out of this part of the test and be somewhere else
tested explictly.

src/testers.jl
```diff
@@ -185,21 +185,13 @@ function frule_test(f, xẋs::Tuple{Any, Any}...; rtol=1e-9, atol=1e-9, fdm=_fdm
     xs, ẋs = first.(xẋs), last.(xẋs)
     Ω_ad, dΩ_ad = frule((NO_FIELDS, ẋs...), f, xs...; fkwargs...)
     Ω = f(xs...; fkwargs...)
-    # if equality check fails, check approximate equality
-    # use collect so can do vector equality
-    # TODO: add isapprox replacement that works for more types
-    @test Ω_ad == Ω || isapprox(collect(Ω_ad), collect(Ω); rtol=rtol, atol=atol)
+    Ω_ad == Ω || isapprox(collect(Ω_ad), collect(Ω); rtol=rtol, atol=atol)
```
oxinabox, line 188:
```suggestion
    @test Ω_ad == Ω || isapprox(collect(Ω_ad), collect(Ω); rtol=rtol, atol=atol)
```

src/testers.jl
```diff
@@ -185,21 +185,13 @@ function frule_test(f, xẋs::Tuple{Any, Any}...; rtol=1e-9, atol=1e-9, fdm=_fdm
...
     # Correctness testing via finite differencing.
     dΩ_fd = _make_jvp_call(fdm, (xs...) -> f(xs...; fkwargs...), xs, ẋs, ẋs_is_ignored)
-    @test isapprox(
-        collect(extern.(dΩ_ad)),  # Use collect so can use vector equality
-        collect(dΩ_fd);
-        rtol=rtol,
-        atol=atol,
-        kwargs...
-    )
+    accumulant = rand_tangent(Ω)
```
oxinabox, line 193: in `frule_test` we don't ask for a accumulant, so her we use the tangent generation stuff to do so.

In `rrule_test` we do ask for an accumulant.

Probably we should actually be asking for an accumulant for both of them
and have some `auto` flag that can be passed in to have one generated via `rand_tangent`.

oxinabox:
I don't really love this PR. though I like it more than the code it replaces i think.
I might sit on if for a few days see if anyone else has good input

src/check_result.jl
```diff
@@ -0,0 +1,51 @@
...
+    @test isapprox(actual, expected; kwargs...)
+end
+
+function check_equal(actual::AbstractThunk, expected; kwargs...)
+    check_equal(unthunk(actual), expected; kwargs...)
+end
+
+
+function check_equal(
+    actual::Union{Composite, AbstractArray},
```
oxinabox, line 26: The `AbstractArray` here is to catch weird things like 
`Vector{Matrix{Float64}}` which we do actually have.

src/testers.jl
sethaxen: Can you add a comment explaining why it's called `accumulant`?

src/testers.jl
```diff
@@ -220,7 +212,7 @@ function rrule_test(f, ȳ, xx̄s::Tuple{Any, Any}...; rtol=1e-9, atol=1e-9, fdm
     _ensure_not_running_on_functor(f, "rrule_test")
 
     # Check correctness of evaluation.
-    xs, x̄s = collect(zip(xx̄s...))
+    xs, accumulated_x̄s = collect(zip(xx̄s...))
```
sethaxen, line 215: Likewise here can you add a comment explaining why it's called `accumulated_`?

src/check_result.jl
```diff
@@ -0,0 +1,51 @@
...
+    expected;
+    kwargs...
+)
+    @test length(actual) == length(expected)
```
sethaxen, line 30: Do we want to check length for `Composite`? e.g. this would fail
```julia
x = (a = 3, b = 4)
dx1 = Composite{typeof(x)}(a = 3, b = Zero())
dx2 = Composite{typeof(x)}(a = 3)
check_equal(dx1, dx2)
```

src/check_result.jl
sethaxen: +1 for more human-readable error messages!

sethaxen (changes requested):
I haven't followed the accumulation PR(s?), so take this review with that caveat. In general I like the introduction of more human-readable errors (can you actually add a test that `check_equal` produces human-readable errors for a case where `isapprox` would not have?), and I don't see a problem with the accumulation part (can you add a test for the testers for an accumulating function?)

src/testers.jl
oxinabox: I think we can leave this for now, I have removed this from the test

src/testers.jl
oxinabox: remove

src/check_result.jl
oxinabox: done

src/testers.jl
```diff
@@ -185,21 +185,13 @@ function frule_test(f, xẋs::Tuple{Any, Any}...; rtol=1e-9, atol=1e-9, fdm=_fdm
     xs, ẋs = first.(xẋs), last.(xẋs)
     Ω_ad, dΩ_ad = frule((NO_FIELDS, ẋs...), f, xs...; fkwargs...)
     Ω = f(xs...; fkwargs...)
-    # if equality check fails, check approximate equality
```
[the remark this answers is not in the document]
oxinabox: whole section is gone

src/check_result.jl
oxinabox: good point

oxinabox:
I have come back to this because i an sick of bad messages around thunks
so it is rebased and has test now 
@sethaxen  can you rereview

Example change:

```
julia> oldcheck(log(-1.0 - im), k)
Test Failed at REPL[95]:1
  Expression: isapprox(actual, expected)
   Evaluated: isapprox(0.34657359027997264 - 2.356194490192345im, Thunk(var"#217#218"()))
ERROR: There was an error during testing

julia> check(log(-1.0 - im), k)
Test Failed at /Users/oxinabox/JuliaEnvs/ChainRulesWorld/ChainRulesTestUtils.jl/src/check_result.jl:18
   Evaluated: isapprox(0.34657359027997264 - 2.356194490192345im, 0.34657359027997264 + 2.356194490192345im)
ERROR: There was an error during testing
```

src/check_result.jl
oxinabox: fixing now

src/testers.jl
```diff
@@ -123,48 +127,40 @@ function test_scalar(f, z; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(),
...
-                rtol=rtol,
-                atol=atol,
-                kwargs...,
-            )
+            _, real_tangent = frule((Zero(), real(Δx)), f, z; fkwargs...)
+            _, embedded_tangent = frule((Zero(), Δx), f, z; fkwargs...)
+            _check_equal(real_tangent, embedded_tangent; isapprox_kwargs...)
         end
     end
     if z isa Complex
```
oxinabox, line 138: I want to move the 4 sections of this function that happen only for complex out into 2 helper functions.
(one for forward on one for reverse)

nickrobinson251:
Does this close https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/7? I think probably, but not sure because i've just given it a quick look.

Also does it make any difference to https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/56?

src/check_result.jl
```diff
@@ -1,3 +1,83 @@
...
+# Note that this must work well both on Differnetial types and Primal types
+
+"""
+    _check_equal(actual, expected; kwargs...)
+
+`@test`'s  that `actual ≈ expected`, but breaks up data such that human readable results
+are shown on failures.
+All keyword arguments are passed to `isapprox`.
+"""
+function _check_equal(
```
oxinabox, line 13: I wonder if we should export this, (and rename it)
for people who need to write special tests?
Might make it feel more safe to remove `isapprox` as there is a handy subsitution

src/check_result.jl
```diff
@@ -1,3 +1,83 @@
...
+        all_keys = union(keys(actual), keys(expected))
+        @testset "$P.$ii" for ii in all_keys
+            _check_equal(getproperty(actual, ii), getproperty(expected, ii); kwargs...)
+        end
+    end
+end
+
+function _check_equal(
+    ::Composite{ActualPrimal}, expected::Composite{ExpectedPrimal}
+    ) where {ActualPrimal, ExpectedPrimal}
```
willtebbutt, line 52:
```suggestion
) where {ActualPrimal, ExpectedPrimal}
```

test/check_result.jl
```diff
@@ -17,4 +17,67 @@
...
     end
+
+
+    @testset "_check_equal" begin
+        check = ChainRulesTestUtils._check_equal
+
+        @testset "possive cases" begin
+            check(1.0, 1.0)
+            check(1.0 + im, 1.0 + im)
+            check(1.0, 1.0+1e-10)  # isapprox _behavour
```
willtebbutt, line 28:
```suggestion
            check(1.0, 1.0+1e-10)  # isapprox _behaviour
```

src/check_result.jl
```diff
@@ -1,3 +1,83 @@
+# For once you have the sensitivity by two methods (e.g  both finite-differencing and  AD)
+# the code here checks it is correct.
+# Goal is to only call `@isapprox` on things that render well
+# Note that this must work well both on Differnetial types and Primal types
```
willtebbutt, line 4:
```suggestion
# Note that this must work well both on Differential types and Primal types
```

src/check_result.jl
```diff
@@ -1,3 +1,83 @@
...
+
+# This catches comparisons of Composites and Tuples/NamedTuple
+# and gives a error messaage complaining about that
+_check_equal(::C, expected::T) where {C<:Composite, T} = @test C === T
+_check_equal(::T, expected::C) where {C<:Composite, T} = @test C === T
+
+
+_check_equal(::AbstractZero, x; kwargs...) = _check_equal(zero(x), x; kwargs...)
+_check_equal(x, ::AbstractZero; kwargs...) = _check_equal(x, zero(x); kwargs...)
+_check_equal(x::AbstractZero, y::AbstractZero; kwargs...) = @test x === y
```
oxinabox, line 66: I am a bit dubious about defining these for `DoesNotExist`
what do you think?

src/iterator.jl
```diff
@@ -51,15 +51,6 @@ end
...
 
-Base.isapprox(iter1::TestIterator, iter2::TestIterator) = false
-function Base.isapprox(
-    iter1::TestIterator{T1,IS,IE},
-    iter2::TestIterator{T2,IS,IE};
-    kwargs...,
-) where {T1,T2,IS,IE}
-    return isapprox(iter1.data, iter2.data; kwargs...)
-end
-
```
oxinabox, line 62: Technically this is breaking, but I am pretty sure they are not used except in this package.
But I could keep them and move them to deprecated, do you think?

src/check_result.jl
willtebbutt: Hmmm good point. I think I agree that it feels a bit weird to define it. What would the practical utility of doing so be?

oxinabox:
> Does this close #7? I think probably, but not sure because i've just given it a quick look.

Yes, this closes #7  (technically would need to wait for next breaking change when `isapprox` on all the differnetials is deleted but its not used internally any more)

>  Also does it make any difference to #56?

#56 was actually closed by the follow up to this PR that was more minimal #59, but I forgot to close it

src/check_result.jl
```diff
@@ -1,3 +1,83 @@
...
```
oxinabox, line 66:
```suggestion
_check_equal(::Zero, x; kwargs...) = _check_equal(zero(x), x; kwargs...)
_check_equal(x, ::Zero; kwargs...) = _check_equal(x, zero(x); kwargs...)
_check_equal(x::Zero, y::Zero; kwargs...) = @test true
```

src/check_result.jl
oxinabox: I have change to just Zero

src/check_result.jl
willtebbutt: As I think we're probably committed to maintaining some functionality of this form, I agree that we should export it.

willtebbutt (approved):

PR #58, opened 2020-10-06
Reorganize test for iterfun

oxinabox:
Very small change to tests,
I worked out why we couldn't put iterfun in the testset.
Its because both used the variable `x` and it was creating a closure.

This fixes it b putting the tests that use `x` in a sub-testset

### DIFF
--- a/test/testers.jl
+++ b/test/testers.jl
@@ -7,38 +7,6 @@ sinconj(x) = sin(x)
 
 primalapprox(x) = x
 
-function iterfun(iter)
-    state = iterate(iter)
-    state === nothing && error()
-    (x, i) = state
-    s = x^2
-    while true
-        state = iterate(iter, i)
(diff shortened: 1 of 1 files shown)


test/testers.jl
```diff
@@ -241,10 +209,44 @@ end
...
+                ∂s += 2 * x * Δx
+            end
+            return s, ∂s
+        end
+
         function ChainRulesCore.rrule(::typeof(iterfun), iter::TestIterator)
             function iterfun_pullback(Δs)
                 data = iter.data
                 ∂data = (2 * Δs) .* conj.(data)
+                @show ∂data
```
nickrobinson251, line 249: do we still want this `@show`? (it wasn't in the code before)

oxinabox: We do not
```suggestion
```

PR #59, opened 2020-10-06
Minimal Inplace accumulation testing

oxinabox:
This is the much more minimal version of  #57  to just test Inplace accumulation.

This is waiting on me building and metatesting tools to catch the negative cases

This is currently built on top of https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/60/files
I will rebase after that is merged

### DIFF
--- a/test/testers.jl
+++ b/test/testers.jl
@@ -42,6 +42,66 @@ primalapprox(x) = x
         end
     end
 
+    @testset "Inplace accumumulation: first on Array" begin
+        @testset "Correct definitions" begin
+            function ChainRulesCore.frule((_, ẋ), ::typeof(first), x::Array)
+                ẏ = InplaceableThunk(
+                    @thunk(first(ẋ)),
+                    ȧ -> ȧ + first(ẋ),  # This won't actually happen inplace
+                )
+                return first(x), ẏ
+            end
+            function ChainRulesCore.rrule(::typeof(first), x::Array{T}) where T
+                x_dims = size(x)
+                function first_pullback(ȳ)
+                    x̄_ret = InplaceableThunk(
+                        Thunk() do
+                            x̄ = zeros(T, x_dims)
+                            x̄[1]=ȳ
+                            x̄
+                        end,
+                        ā -> (ā[1] += ȳ; ā)
+                    )
+                    return (NO_FIELDS, x̄_ret)
+                end
+                return first(x), first_pullback
+            end
+
+            frule_test(first, (randn(4), randn(4)))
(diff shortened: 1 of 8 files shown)


test/testers.jl
```diff
@@ -42,6 +42,66 @@ primalapprox(x) = x
         end
     end
 
+    @testset "Inplace accumumulation: first on Array" begin
+        @testset "Correct definitions" begin
+            function ChainRulesCore.frule((_, ẋ), ::typeof(first), x::Array)
+                ẏ = InplaceableThunk(
+                    @thunk(first(ẋ)),
+                    ȧ -> ȧ + first(ẋ),  # This won't actually happen inplace
```
oxinabox, line 50: Its hard to find forward-mode rules that are both simple and can be done in-place.
(first on a array of arrays is one of them i guess)

But i don't think we actually need something inplace to test this.

The real test of this functionality in detail is in the test for `check_result.jl`

test/testers.jl
```diff
@@ -42,6 +42,66 @@ primalapprox(x) = x
...
+                end
+                return first(x), first_pullback
+            end
+
+            frule_test(first, (randn(4), randn(4)))
+            rrule_test(first, randn(), (randn(4), randn(4)))
+        end
+
+        @testset "Incorrect inplace definitions" begin
+            bad_first(value) = first(value)
```
oxinabox, line 75: `bad_first` isn't bad, but the sensitivities we define for it are.
should i name it something else?

test/testers.jl
```diff
@@ -42,6 +42,66 @@ primalapprox(x) = x
...
+                        end,
+                        ā -> (ā[1] += 1000*ȳ; ā)  # incorrect
+                    )
+                    return (NO_FIELDS, x̄_ret)
+                end
+                return first(x), badfirst_pullback
+            end
+
+            @test fails(()->frule_test(bad_first, (randn(4), randn(4))))
+            @test fails(()->rrule_test(bad_first, randn(), (randn(4), randn(4))))
```
oxinabox, line 100: I manually inspected the output of these to make sure htey were indeed failing in the way we expected

test/testers.jl
nickrobinson251: just `_first`? or `my_first`?

nickrobinson251 (approved):
seems right to me

oxinabox:
I will validate that the Integration test failures are not false positives before merging this

oxinabox:
Will merge this once CI passes

PR #60, opened 2020-10-06
Test the unhappy path

oxinabox:
This PR is about testing that we can actually detect when things are wrong.

The code for `metatest_get_failures` is a pretty nasty hack an on the internals.
I think its clear enough what it is doing.

Two possible alternative implementations:

### Alt 1: use a seperate process.
Julia itself does this to test `@test`.
Writing a test to a file and then kicking it off with ``run(`julia -e file.jl`)``
But three problems are

1. Slow starting an external process is slow
2. Very course grained: all that is easy to test is that it caused julia to to exit with a nonzero status
3. Pretty ugly code, need to redirect stdout still and need to write a file

### Alt 2: custom Testset type
We could declare a custom testset type which overloads `record` and `finish` so that they don't propate errors upwards.
This might be fiddly, but I have done something like it before in https://github.com/JuliaTesting/ReferenceTests.jl.
so could base it off that.
It should be simplier than that.
It could be pretty clean maybe. Not 100% sure.

### DIFF
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -9,24 +9,24 @@ using Test: DefaultTestSet
     nonpassing_results(f)
 
 `f` should be a function that takes no argument, and calls some code that used `@test`.
-Invoking it via `metatest_get_failures(f)` will prevent those `@test` being added to the
+Invoking it via `nonpassing_results(f)` will prevent those `@test` being added to the
 current testset, and will return a collection of all nonpassing test results.
 """
 function nonpassing_results(f)
     mute() do
-        failures = []
+        nonpasses = []
         # Specify testset type incase parent testset is some other typer
         @testset DefaultTestSet "nonpassing internal" begin
             f()
             ts = Test.get_testset()  # this is the current testset "nonpassing internal"
-            failures = _extract_nonpasses(ts)
+            nonpasses = _extract_nonpasses(ts)
             # Prevent the failure being recorded in parent testset.
             empty!(ts.results)
             ts.anynonpass = false
         end
         # Note: we allow the "nonpassing internal" testset to still be pushed as an empty
         # passing testset in its parent testset. We could remove that if we wanted
-        return failures
+        return nonpasses
     end
 end
 
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -14,7 +14,7 @@ function metatest_get_failures(f)
         failures = []
         @testset "dummy" begin
             f()
-            ts = Test.get_testset()
+            ts = Test.get_testset()  # this is the current testset "dummy"
             failures = _extract_failures(ts)
             # Prevent the failure being recorded in parent testset.
             empty!(ts.results)
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -10,18 +10,22 @@ Invoking it via `metatest_get_failures(f)` will prevent those `@test` being adde
 current testset, and will return a collection of all nonpassing test results.
 """
 function metatest_get_failures(f)
-    redirect_stdout(devnull) do
-        failures = []
-        @testset "dummy" begin
-            f()
-            ts = Test.get_testset()  # this is the current testset "dummy"
-            failures = _extract_failures(ts)
-            # Prevent the failure being recorded in parent testset.
-            empty!(ts.results)
-            ts.anynonpass = false
-        end
-        return failures
-   end
+    # TODO: once we are on Julia 1.6 this can be change to just use
+    # `redirect_stdout(devnull)` See: https://github.com/JuliaLang/julia/pull/36146
+    mktemp() do path, tempio
+        redirect_stdout(tempio) do 
+            failures = []
+            @testset "dummy" begin
+                f()
+                ts = Test.get_testset()  # this is the current testset "dummy"
+                failures = _extract_failures(ts)
+                # Prevent the failure being recorded in parent testset.
+                empty!(ts.results)
+                ts.anynonpass = false
+            end
+            return failures
+       end
+    end
 end
 
 "extracts as flat collection of failures from a (potential nested) testset"
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -10,7 +10,7 @@ Invoking it via `metatest_get_failures(f)` will prevent those `@test` being adde
 current testset, and will return a collection of all nonpassing test results.
 """
 function metatest_get_failures(f)
-    redirect_stdout(stdout) do
+    redirect_stdout(devnull) do
         failures = []
         @testset "dummy" begin
             f()
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -1,0 +1,71 @@
+# This is tools for testing ChainRulesTestUtils itself
+# if they were less nasty in implementation we might consider moving them to a package
+# MetaTesting.jl
+
+"""
+    metatest_get_failures(f)
+
+`f` should be a function that takes no argument, and calls some code that used `@test`.
+Invoking it via `metatest_get_failures(f)` will prevent those `@test` being added to the
+current testset, and will return a collection of all nonpassing test results.
+"""
+function metatest_get_failures(f)
+    redirect_stdout(stdout) do
+        failures = []
+        @testset "dummy" begin
+            f()
+            ts = Test.get_testset()
+            failures = _extract_failures(ts)
+            # Prevent the failure being recorded in parent testset.
+            empty!(ts.results)
+            ts.anynonpass = false
+        end
+        return failures
+   end
+end
+
+"extracts as flat collection of failures from a (potential nested) testset"
+_extract_failures(x::Test.Result) = [x,]
+_extract_failures(x::Test.Pass) = Test.Result[]
+_extract_failures(ts::Test.DefaultTestSet) = _extract_failures(ts.results)
+function _extract_failures(xs::Vector)
+    if isempty(xs)
+        return Test.Result[]
+    else
+        return mapreduce(_extract_failures, vcat, xs)
+    end
+end
+
+#Meta Meta tests
+@testset "meta_testing_tools.jl" begin
+    @testset "metatest_get_failures" begin
+        @testset "No Tests" begin
+            fails = metatest_get_failures(()->nothing)
+            @test length(fails) === 0
+        end
+
+        @testset "No Failures" begin
+            fails = metatest_get_failures(()->@test true)
+            @test length(fails) === 0
+        end
+
+
+        @testset "Single Test" begin
+            fails = metatest_get_failures(()->@test false)
+            @test length(fails) === 1
+            @test fails[1].orig_expr == false
+        end
+
+        @testset "Single Testset" begin
+            fails = metatest_get_failures() do
+                @testset "inner" begin
+                    @test false == true
+                    @test true == false
+                end
+            end
+            @test length(fails) === 2
+            @test fails[1].orig_expr == :(false==true)
+            @test fails[2].orig_expr == :(true==false)
+        end
+    end
+end
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -2,66 +2,105 @@
 # if they were less nasty in implementation we might consider moving them to a package
 # MetaTesting.jl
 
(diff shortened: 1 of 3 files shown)


test/meta_testing_tools.jl
```diff
@@ -0,0 +1,71 @@
...
+# MetaTesting.jl
+
+"""
+    metatest_get_failures(f)
+
+`f` should be a function that takes no argument, and calls some code that used `@test`.
+Invoking it via `metatest_get_failures(f)` will prevent those `@test` being added to the
+current testset, and will return a collection of all nonpassing test results.
+"""
+function metatest_get_failures(f)
```
oxinabox, line 12: Should this just be called
```suggestion
function get_failures(f)
```
?

test/testers.jl
```diff
@@ -42,9 +42,17 @@ end
...
+        @testset "happy path" begin
+            double(x) = 2x
+            @scalar_rule(double(x), 2)
+            test_scalar(double, 2.1)
+        end
+        @testset "unhappy path" begin
+            alt_double(x) = 2x
+            @scalar_rule(alt_double(x), 3)  # this is wrong, on purpose
+            failures = metatest_get_failures(()->test_scalar(alt_double, 2.1))
+            @test !isempty(failures)
```
oxinabox, line 54: Should this, and the ones below do more than just `@test !isempty(failures)`?
Like maybe test that the failure is the one we expect?
Or is this good for now?
I worry testing exact failure condition might be too brittle.

test/meta_testing_tools.jl
nickrobinson251: yeah, `extract_failures` or `fail_results` (since they are `Test.Fail <: Test.Result`) or just  `failures`?

test/meta_testing_tools.jl
```diff
@@ -0,0 +1,75 @@
+# This is tools for testing ChainRulesTestUtils itself
+# if they were less nasty in implementation we might consider moving them to a package
+# MetaTesting.jl
```
nickrobinson251, line 3: yeah, i see why this is helpful to us in CRTestUtils, but i think down-the-road we want to move this to it's own package, since it's not CRTestUtils specific :)

test/testers.jl
nickrobinson251: i think this is fine for now :)

test/testers.jl
```diff
@@ -263,4 +271,42 @@ end
...
+                return 2.5*x, ẏ
+            end
+            function ChainRulesCore.rrule(::typeof(my_identity1), x)
+                function identity_pullback(ȳ)
+                    return (NO_FIELDS, ȳ)
+                end
+                return 2.5 * x, identity_pullback
+            end
+
+            @test !isempty(metatest_get_failures(()->frule_test(my_identity1, (2.2, 3.3))))
```
nickrobinson251, line 288: I do wonder if there's an alternative design (or just another layer of abstraction?)
so we end up with a call more like
```julia
@test_failure frule_test(my_identity1, (2.2, 3.3))
```
or
```
@test fails(() -> frule_test(my_identity1, (2.2, 3.3))
```
?

test/testers.jl
```diff
@@ -263,4 +271,42 @@ end
         frule_test(iterfun, (x, ẋ))
         rrule_test(iterfun, randn(), (x, x̄))
     end
+
+    @testset "unhappy path" begin
+        @testset "primal wrong" begin
+            my_identity1(x) = x
+            function ChainRulesCore.frule((_, ẏ), ::typeof(my_identity1), x)
+                return 2.5*x, ẏ
```
nickrobinson251, line 279:
```suggestion
                return 2.5 * x, ẏ
```

test/testers.jl
```diff
@@ -42,9 +42,17 @@ end
...
-        test_scalar(double, 2.0)
+        @testset "happy path" begin
+            double(x) = 2x
+            @scalar_rule(double(x), 2)
+            test_scalar(double, 2.1)
+        end
+        @testset "unhappy path" begin
+            alt_double(x) = 2x
+            @scalar_rule(alt_double(x), 3)  # this is wrong, on purpose
+            failures = metatest_get_failures(()->test_scalar(alt_double, 2.1))
```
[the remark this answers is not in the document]
nickrobinson251: i wonder if the testset name can be more precise on that, than `"unhappy path"`

nickrobinson251 (commented):
(reviewed with @bencottier)

one design question, and a couple naming comments, but otherwise i think the broadly implementation looks good -- i've asked Ben to follow-up with any comments on the implementation later this afternoon

test/meta_testing_tools.jl
oxinabox: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/62

test/meta_testing_tools.jl
oxinabox: They are any non-pass.
So they could be Errors, or some other cause, I don't recall what `@test_broken` returns.
Arguably we should call it `nonpasses`

test/testers.jl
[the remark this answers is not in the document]
oxinabox: we want to make sure it doesn't succeed.

I can't think of a better name.
Unhappy Path is a term of literature
https://cucumber.io/blog/test-automation/happy-unhappy-paths-why-you-need-to-test-both/

test/testers.jl
oxinabox: Cool, can always make it tighter if something slips through later

test/testers.jl
oxinabox: Yeah i think the later makes sense.

oxinabox:
@nickrobinson251  you didn't comment on if either of the Alternative Implementations mentioned in the top  post might be better.

nickrobinson251:
> @nickrobinson251 you didn't comment on if either of the Alternative Implementations mentioned in the top post might be better.

We did. We think Alt 1 is clearly worse. Alt 2 we didn't think too hard about but I couldn't see what we'd gain over the current implementation

oxinabox:
Ah i missed it

> overall Alt 1 and (probably) Alt 2 both seem less nice to us

test/testers.jl
nickrobinson251: out of interest, why not the macro (that just wraps the call in an anonymous function and passes it to `fails`)?

test/meta_testing_tools.jl
```diff
@@ -0,0 +1,75 @@
...
+                ts.anynonpass = false
+            end
+            return failures
+       end
+    end
+end
+
+"extracts as flat collection of failures from a (potential nested) testset"
+_extract_failures(x::Test.Result) = [x,]
+_extract_failures(x::Test.Pass) = Test.Result[]
```
nickrobinson251, line 33: don't we only want `Test.Fail`s to be extracted?

right now, it seems `Fail` and `Error` (and everything that's not `Pass`) get treated the same - is that right - why is that?

test/testers.jl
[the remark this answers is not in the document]
nickrobinson251: > we want to make sure it doesn't succeed.

i am confused by this -- why don't we only care about failures? 
(also current naming is confusing to me if it includes more than `Test.Fail`s)
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/60/files#r501079347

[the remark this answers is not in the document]
nickrobinson251: this is also why i prefer to be more precise than "unhappy path" ...unless we explicitly document that "unhappy path" means "anything that doesn't result in only `Test.Pass`"

test/meta_testing_tools.jl
oxinabox: I have renamed to refer to non-passes and made `fail`s count it specially.

test/meta_testing_tools.jl
nickrobinson251: yeah `nonpasses` is better (to avoid the `failures` = `Fail` confusion i had)

`test_broken` also treats all non-passes the same
```julia
julia> @test_broken 1 == 2
Test Broken
  Expression: 1 == 2

julia> @test_broken error()
Test Broken
  Expression: error()
```

test/meta_testing_tools.jl
nickrobinson251: sounds good!

i don't see those commits yet -- did you push them?

test/testers.jl
oxinabox: the macro looks like `@test_throws` and `@test_broken`
but it does not at all work like those do.

`@test_throws` catchs the exception.
`@test_broken x` is very similar to `@test !x` + some extra stuff

I would worry that people might think  `@test_failure x`  is the same as same as `@test !x`

test/meta_testing_tools.jl
oxinabox: Not yet, made them as the RSE meetign was about to start

test/meta_testing_tools.jl
```diff
@@ -0,0 +1,75 @@
...
+_extract_failures(ts::Test.DefaultTestSet) = _extract_failures(ts.results)
+function _extract_failures(xs::Vector)
+    if isempty(xs)
+        return Test.Result[]
```
bencottier, line 37: If `f` actually doesn't call `@test`, we end up here -- is this the best thing to return in that case?

test/testers.jl
nickrobinson251: > people might think @test_failure x is the same as same as @test !x

yeah, great point.

test/meta_testing_tools.jl
oxinabox: I think so?

test/meta_testing_tools.jl
```diff
@@ -0,0 +1,132 @@
...
+# MetaTesting.jl
+
+# need to bring this into scope explictly so can use in @testset nonpassing_results
+using Test: DefaultTestSet
+
+"""
+    nonpassing_results(f)
+
+`f` should be a function that takes no argument, and calls some code that used `@test`.
+Invoking it via `metatest_get_failures(f)` will prevent those `@test` being added to the
```
nickrobinson251, line 12:
```suggestion
Invoking it via `nonpassing_results(f)` will prevent those `@test` being added to the
```

test/meta_testing_tools.jl
```diff
@@ -0,0 +1,132 @@
...
+current testset, and will return a collection of all nonpassing test results.
+"""
+function nonpassing_results(f)
+    mute() do
+        failures = []
```
nickrobinson251, line 17:
```suggestion
        nonpasses = []
```

test/meta_testing_tools.jl
```diff
@@ -0,0 +1,132 @@
...
+        # Specify testset type incase parent testset is some other typer
+        @testset DefaultTestSet "nonpassing internal" begin
+            f()
+            ts = Test.get_testset()  # this is the current testset "nonpassing internal"
+            failures = _extract_nonpasses(ts)
```
nickrobinson251, line 22:
```suggestion
            nonpasses = _extract_nonpasses(ts)
```

test/meta_testing_tools.jl
```diff
@@ -0,0 +1,132 @@
...
+            # Prevent the failure being recorded in parent testset.
+            empty!(ts.results)
+            ts.anynonpass = false
+        end
+        # Note: we allow the "nonpassing internal" testset to still be pushed as an empty
+        # passing testset in its parent testset. We could remove that if we wanted
+        return failures
```
nickrobinson251, line 29:
```suggestion
        return nonpasses
```

nickrobinson251 (approved):
💚

test/testers.jl
```diff
@@ -263,4 +270,38 @@ end
...
+            end
+            function ChainRulesCore.rrule(::typeof(my_identity1), x)
+                function identity_pullback(ȳ)
+                    return (NO_FIELDS, ȳ)
+                end
+                return 2.5 * x, identity_pullback
+            end
+
+            @test fails(()->frule_test(my_identity1, (2.2, 3.3))))
+            @test fails(()->rrule_test(my_identity1, 4.1, (2.2, 3.3))))
```
oxinabox, line 288:
```suggestion
            @test fails(()->frule_test(my_identity1, (2.2, 3.3)))
            @test fails(()->rrule_test(my_identity1, 4.1, (2.2, 3.3)))
```

oxinabox:
Not going to tag a release off this as it is not a functional change

Issue #63, opened 2020-10-08
Migrate to new FiniteDifferences v0.11


oxinabox:
We should move to FiniteDifferences v0.11

PR #65, opened 2020-10-09
Update FiniteDifferences to v0.11

bencottier:
#63

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,5 +13,5 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 [compat]
 ChainRulesCore = "0.9.1"
 Compat = "3"
-FiniteDifferences = "0.10"
+FiniteDifferences = "0.11"
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.5.2"
+version = "0.5.3"
 
 [deps]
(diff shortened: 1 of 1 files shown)


bencottier:
It looks like `ChainRules` also needs to migrate for IntegrationTestChainRules to pass.

oxinabox:
> It looks like ChainRules also needs to migrate for IntegrationTestChainRules to pass.

Yep, ideally ChainRules would not use FiniteDIfferences.jl directly.
only via ChainRulesTestUtils, but we don't live in that ideal world.
I don't remember why not.
It might be that ChainRulesTestUtils was missing a feature, that has since been added.

oxinabox:
I am *really* surprised that this didn't break anything.
Or even throw any deprecation warnings

Project.toml
```diff
@@ -13,5 +13,5 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 [compat]
 ChainRulesCore = "0.9.1"
 Compat = "3"
-FiniteDifferences = "0.10"
+FiniteDifferences = "0.11"
```
oxinabox, line 16: Since this didn't require any code changes,
we can keep supporting FiniteDifferences 0.10 as well.
So

```suggestion
FiniteDifferences = "0.10, 0.11"
```

oxinabox (approved):
Can you also bump the version number, so we can tag a release?
This is a nonbreaking change,
so just incrementing the final version number

oxinabox (approved):
LGTM,
merging

mzgubic:
> #63

If you do "Closes #63" the issue will be automatically closed once the MR is merged🧙

PR #70, opened 2020-10-15
Fix tests now that add!! is too smart for them

oxinabox:
This is a follow up to https://github.com/JuliaDiff/ChainRulesCore.jl/pull/234
but it doesn't need that PR to be merged first,
as it is a valid test both before and after that change.
The test before was being lazy and once https://github.com/JuliaDiff/ChainRulesCore.jl/pull/234 goes in will start failing

### DIFF
--- a/test/testers.jl
+++ b/test/testers.jl
@@ -44,20 +44,26 @@ primalapprox(x) = x
 
     @testset "Inplace accumulation: identity on Array" begin
         @testset "Correct definitions" begin
+            local inplace_used
             function ChainRulesCore.frule((_, ẋ), ::typeof(identity), x::Array)
-                ẏ = InplaceableThunk(@thunk(ẋ), ȧ -> ȧ .+= ẋ)
+                ẏ = InplaceableThunk(@thunk(ẋ), ȧ -> (inplace_used=true; ȧ .+= ẋ))
                 return identity(x), ẏ
             end
             function ChainRulesCore.rrule(::typeof(identity), x::Array)
                 function identity_pullback(ȳ)
-                    x̄_ret = InplaceableThunk(@thunk(ȳ), ā -> ā .+= ȳ)
+                    x̄_ret = InplaceableThunk(@thunk(ȳ), ā -> (inplace_used=true; ā .+= ȳ))
                     return (NO_FIELDS, x̄_ret)
                 end
                 return identity(x), identity_pullback
             end
 
+            inplace_used = false
             frule_test(identity, (randn(4), randn(4)))
+            @test inplace_used  # make sure we are using, and thus testing the add!
+
+            inplace_used = false
             rrule_test(identity, randn(4), (randn(4), randn(4)))
+            @test inplace_used  # make sure we are using, and thus testing the add!
         end
 
         @testset "Incorrect in-place definitions" begin
--- a/test/testers.jl
+++ b/test/testers.jl
@@ -50,7 +50,7 @@ primalapprox(x) = x
             end
             function ChainRulesCore.rrule(::typeof(identity), x::Array)
                 function identity_pullback(ȳ)
-                    x̄_ret = InplaceableThunk(@thunk(x̄), ā -> ā .+= ȳ)
+                    x̄_ret = InplaceableThunk(@thunk(ȳ), ā -> ā .+= ȳ)
                     return (NO_FIELDS, x̄_ret)
                 end
                 return identity(x), identity_pullback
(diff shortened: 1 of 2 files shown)


test/testers.jl
```diff
@@ -42,62 +42,42 @@ primalapprox(x) = x
...
-                        end,
-                        ā -> (ā[1] += ȳ; ā)
-                    )
+            function ChainRulesCore.rrule(::typeof(identity), x::Array)
+                function identity_pullback(ȳ)
+                    x̄_ret = InplaceableThunk(@thunk(x̄), ā -> ā .+= ȳ)
                     return (NO_FIELDS, x̄_ret)
                 end
-                return first(x), first_pullback
+                return identity(x), first_pullback
```
oxinabox, line 56:
```suggestion
                return identity(x), identity_pullback
```

test/testers.jl
```diff
@@ -42,62 +42,42 @@ primalapprox(x) = x
...
                     return (NO_FIELDS, x̄_ret)
                 end
-                return first(x), first_pullback
+                return identity(x), identity_pullback
             end
 
-            frule_test(first, (randn(4), randn(4)))
-            rrule_test(first, randn(), (randn(4), randn(4)))
+            frule_test(identity, (randn(4), randn(4)))
+            rrule_test(identity, randn(4), (randn(4), randn(4)))
```
mzgubic, line 60: I am not sure I understand what this test was meant to catch before (and therefore how the changes improve it).

Was it just a general test of inplace accummulation using `InplaceableThunk`s? If so, does the new `add!!` function just use the normal `x+y` rather than the inplace computation?

test/testers.jl
```diff
@@ -42,62 +42,42 @@ primalapprox(x) = x
         end
     end
 
-    @testset "Inplace accumumulation: first on Array" begin
+    @testset "Inplace accumulation: identity on Array" begin
         @testset "Correct definitions" begin
-            function ChainRulesCore.frule((_, ẋ), ::typeof(first), x::Array)
-                ẏ = InplaceableThunk(
-                    @thunk(first(ẋ)),
-                    ȧ -> ȧ + first(ẋ),  # This won't actually happen inplace
```
mzgubic, line 50: I guess an outdated comment, but that is only true for the current input, right? (i.e. `rand(4)`).

My understanding it that `first` would not make a copy of an array, because of this:
```
julia> a = rand(3)
3-element Array{Float64,1}:
 0.36709403077151537
 0.49246683838177385
 0.8083301928782081 

julia> A = [a, rand(3)]
2-element Array{Array{Float64,1},1}:
 [0.36709403077151537, 0.49246683838177385, 0.8083301928782081]
 [0.822528854921962, 0.6952904499081514, 0.43481292252062986]  

julia> first(A) === a
true
```
But maybe this isn't testing what I think it is testing.

test/testers.jl
```diff
@@ -42,62 +42,42 @@ primalapprox(x) = x
...
-            function ChainRulesCore.frule((_, ẋ), ::typeof(my_first), x::Array)
-                ẏ = InplaceableThunk(
-                    @thunk(first(ẋ)),  # correct
-                    ȧ -> ȧ + 1000*first(ẋ),  # incorrect (also not actually inplace)
-                )
-                return first(x), ẏ
+        @testset "Incorrect in-place definitions" begin
+            my_identity(value) = value  # we will define bad rules on this
+            function ChainRulesCore.frule((_, ẋ), ::typeof(my_identity), x::Array)
+                # only the in-place part is incorrect
```
mzgubic, line 66: I guess the point is that we are checking whether inplace accumulation actually happens?

The test succeeds if the result is unthunked, but not if inplace accumulation is used?

test/testers.jl
```diff
@@ -23,12 +23,12 @@ primalapprox(x) = x
...
-        function ChainRulesCore.frule((_, ẏ), ::typeof(identity), x)
-            return x, ẏ
+        function ChainRulesCore.frule((_, ẏ), ::typeof(identity), x)
+            return x, ẏ
         end
         function ChainRulesCore.rrule(::typeof(identity), x)
-            function identity_pullback(ȳ)
-                return (NO_FIELDS, ȳ)
+            function identity_pullback(ȳ)
+                return (NO_FIELDS, ȳ)
```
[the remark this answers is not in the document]
oxinabox: I think maybe github bugged out? Or some white space was removed

test/testers.jl
```diff
@@ -42,62 +42,42 @@ primalapprox(x) = x
         end
     end
 
-    @testset "Inplace accumumulation: first on Array" begin
+    @testset "Inplace accumulation: identity on Array" begin
```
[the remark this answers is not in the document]
oxinabox: because first can't be done inplace for forward mode

test/testers.jl
oxinabox: Yes, just a general trest of inplace accumulation using `InplacableThunk`.
But the new `add!!` is too smart and will do `x+y` instread of the inplace computation,
thus defeating the point of this test.
It's not a great test in that it can be silently defeated.
Ideally I guess we would intrument it and check that the code paths we want to be hit are hit.
I think i can workout how to do that.

test/testers.jl
oxinabox: the reason it isn't acxtually happening inplace is because because it doesn't mutate the content of `ȧ`

test/testers.jl
oxinabox: correct

mzgubic (approved):

mzgubic:
the codepath checks lgtm

PR #72, opened 2020-10-19
stringify orig_expr in meta-metatests

oxinabox:
Julia broke this on current master
https://github.com/JuliaLang/julia/pull/37809#issuecomment-712162789

I am not particularly in a hurry to see this merged.
I don't love this change, since it is now sensitive to the spacing that julia happens to use for `==`

### DIFF
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -96,7 +96,10 @@ end
         @testset "Single Test" begin
             fails = nonpassing_results(()->@test false)
             @test length(fails) === 1
-            @test fails[1].orig_expr == false
+            # Julia 1.6 return a `String`, not an `Expr`.
+            # Always calling  `string` on it gives gives consistency regardless of version.
+            # https://github.com/JuliaLang/julia/pull/37809
+            @test string(fails[1].orig_expr) == string(false)
         end
 
         @testset "Single Testset" begin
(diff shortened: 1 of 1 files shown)


test/meta_testing_tools.jl
```diff
@@ -107,8 +107,13 @@ end
...
             @test length(fails) === 2
-            @test fails[1].orig_expr == :(false==true)
-            @test fails[2].orig_expr == :(true==false)
+            
+
+            # Newer versions of Julia return a `String`, not an `Expr`. 
+            # Always calling  `string` on it gives gives consistency regardless of version.
+            # https://github.com/JuliaLang/julia/pull/37809
+            @test string(fails[1].orig_expr) == "false == true"
+            @test string(fails[2].orig_expr) == "true == false"
```
nickrobinson251, line 116: alternatively we could check if `orig_expr` is a `String` (or check the Julia version) and `Meta.parse` it to an `Expr`?

nickrobinson251: then we'd be testing what we want to test (which is that the expression is the one we expect), rather than testing the string representation

oxinabox: indeed, but it is more complex

oxinabox: I still lean towards simpler is better

mzgubic: I've initially done what @nickrobinson251 suggested but I prefer this. In particular the current one which stringifies the expression

mzgubic (approved):

PR #73, opened 2020-11-08
add support of 0-dim array for `rand_tangent`

ho-oto:
This patch fixes
```julia
a = fill(1.0) # typeof(a) == Array{Float64,0}
b = rand_tangent(a) # typeof(b) == Float64
typeof(a) == typeof(b) # => false
```

### DIFF
--- a/test/generate_tangent.jl
+++ b/test/generate_tangent.jl
@@ -18,7 +18,7 @@ end
         (4, DoesNotExist),
         (5.0, Float64),
         (5.0 + 0.4im, Complex{Float64}),
-        (fill(6.0), Array{Float64,0}),
+        (fill(6.0), Array{Float64, 0}),
         (randn(Float32, 3), Vector{Float32}),
         (randn(Complex{Float64}, 2), Vector{Complex{Float64}}),
         (randn(5, 4), Matrix{Float64}),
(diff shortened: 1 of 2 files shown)


willtebbutt:
Hmm interesting. I wonder if there's a way to refactor the existing implementation so that it generalised nicely to this case.

Looks like if we change the `broadcast` to `map` in the `rand_tangent(rng::AbstractRNG, x::StridedArray)` method we get the correct behaviour. @ho-oto could you confirm that this also works for you?

ho-oto:
Yes. It works for me. Should I update my patch?

willtebbutt:
Yes please :) Will be nice to have this resolved.

test/generate_tangent.jl
```diff
@@ -18,6 +18,7 @@ end
         (4, DoesNotExist),
         (5.0, Float64),
         (5.0 + 0.4im, Complex{Float64}),
+        (fill(6.0), Array{Float64,0}),
```
willtebbutt, line 21:
```suggestion
        (fill(6.0), Array{Float64, 0}),
```

willtebbutt (approved):
Minor style thing, but given that and a patch version bump, I'm happy. Thanks for the contribution!

nickrobinson251:
Would you mind reading this to resolve conflicts? Then hopefully we can merge it :)

oxinabox:
This code has moved to FiniteDifferences.jl so nothing left to do here.
Would be good to see if the issue still exists there and if so a new PR could be made?

PR #75, opened 2020-11-28
Implement isapprox for Composites

willtebbutt:
What it says on the tin.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.5.3"
+version = "0.5.4"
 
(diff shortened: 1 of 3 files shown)


mzgubic (approved):

willtebbutt:
Thanks for approving @mzgubic . Should I be worried about the integration test failure?

mzgubic:
> Thanks for approving @mzgubic . Should I be worried about the integration test failure?

I have seen it on a few recent PRs but could not reproduce this failure on my machine. Perhaps we should investigate? It seems like there is something wrong with CI but this is not really my area of expertise.

willtebbutt:
Ah okay. Maybe open an issue about it? I'll merge this in the mean time.

mzgubic:
will do

PR #78, opened 2020-12-05
Add option to check if frules and rrules are type-inferrable

sethaxen:
As discussed with @willtebbutt in https://github.com/JuliaDiff/ChainRules.jl/pull/321#issuecomment-739079403, this PR adds a `check_inferred=true` default keyword argument to the tester functions to check if the frules, rrules, pullbacks, and thunks are all type-stable.

The functionality is there, but I'm having some difficulty testing that `test_scalar` fails when an `rrule` or `frule` is type-unstable. Perhaps a testing kung fu master could give me a pointer.


willtebbutt:
Hmmm I'm stumped by this one -- not a massive `Test.jl` expert though. Maybe someone on slack or discourse could assist?

oxinabox:
> The functionality is there, but I'm having some difficulty testing that test_scalar fails when an rrule or frule is type-unstable. Perhaps a testing kung fu master could give me a pointer.

We have a scary, but functional tool for this in
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/master/test/meta_testing_tools.jl

Basically do `@test fails(()-><<CODE THAT WILL CALLS A TEST FAILURE>>)`
you can see some examples in that file

src/testers.jl
```diff
@@ -248,6 +269,8 @@ function rrule_test(f, ȳ, xx̄s::Tuple{Any, Any}...; rtol=1e-9, atol=1e-9, fdm
             @assert x̄_fd === nothing  # this is how `_make_j′vp_call` works
             @test x̄_ad isa DoesNotExist  # we said it wasn't differentiable.
         else
+            x̄_ad isa Thunk && check_inferred && _test_inferred(unthunk, x̄_ad)
```
oxinabox, line 272:
```suggestion
            x̄_ad isa AbstractThunk && check_inferred && _test_inferred(unthunk, x̄_ad)
```

oxinabox (approved):
Other than missing the unhappy path tests, this seems good.
It is a breaking change, since it defaults to true, but I think that is fine?

sethaxen:
> It is a breaking change, since it defaults to true, but I think that is fine?

Yes, I think we do want it to be on by default. It is already exposing a fair amount of type-instability in ChainRules. It _is_ a breaking change, and it needs a bump to minor version number.

sethaxen:
This is ready to merge, but because it bumps the minor version number and will require some changes to ChainRules to bring back compatibility, maybe I should wait for other non-breaking PRs to be merged first?

willtebbutt:
I have no objection to it being merged, but @oxinabox might want to get a couple of his things in first?

oxinabox:
yes, I want to get #83 and #84 in first.
and also can we leave this as `-DEV` as there are some deprecations to be remvoved also

oxinabox:
This can be merged now

PR #83, opened 2020-12-15
support isapprox still

oxinabox:
fixed https://github.com/invenia/BlockDiagonals.jl/issues/49
which was overloading `isapprox` to support compare natural and structural differentials
which was broken by https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/57/

I want to do more to support those without them overloading `isapprox`

### DIFF
--- a/src/check_result.jl
+++ b/src/check_result.jl
@@ -27,7 +27,8 @@ end
 
 """
     _can_pass_early(actual, expected; kwargs...)
-Used to check if we actual is basically equal to expected, so we don't need to check deeper.
+Used to check if `actual` is basically equal to `expected`, so we don't need to check deeper;
+and can just report `check_equal` as passing.
 
 If either `==` or `≈` return true then so does this.
 The `kwargs` are passed on to `isapprox`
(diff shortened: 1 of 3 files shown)


src/check_result.jl
```diff
@@ -25,9 +25,29 @@ for (T1, T2) in ((AbstractThunk, Any), (AbstractThunk, AbstractThunk), (Any, Abs
     end
 end
 
+"""
+    _can_pass_early(actual, expected; kwargs...)
+Used to check if we actual is basically equal to expected, so we don't need to check deeper.
```
mzgubic, line 30:
```suggestion
Used to check if the actual is basically equal to expected, so we don't need to check deeper.
```

mzgubic (approved):
Took me a while to figure out how this all fits together, but LGTM

src/check_result.jl
oxinabox:
```suggestion
Used to check if `actual` is basically equal to `expected`, so we don't need to check deeper;
and can just report `check_equal` as passing.
```

PR #84, opened 2020-12-15
check_equality on structural x natural

oxinabox:
Follow up from #83 that removes the need for `isapprox` to be defined.
This should be merged *after* #83 
this PR was cut onto of that one and only the last commit is new

### DIFF
--- a/src/check_result.jl
+++ b/src/check_result.jl
@@ -96,7 +96,7 @@ function check_equal(actual::Composite{P, T}, expected; kwargs...) where {T, P}
         end
     end
 end
-check_equal(x, y::Composite; kwargs...) = check_equal(y, x; args...)
+check_equal(x, y::Composite; kwargs...) = check_equal(y, x; kwargs...)
 
 # This catches comparisons of Composites and Tuples/NamedTuple
 # and gives a error messaage complaining about that
--- a/src/check_result.jl
+++ b/src/check_result.jl
@@ -25,6 +25,10 @@ for (T1, T2) in ((AbstractThunk, Any), (AbstractThunk, AbstractThunk), (Any, Abs
     end
 end
 
+check_equal(::Zero, x; kwargs...) = check_equal(zero(x), x; kwargs...)
+check_equal(x, ::Zero; kwargs...) = check_equal(x, zero(x); kwargs...)
+check_equal(x::Zero, y::Zero; kwargs...) = @test true
+
 """
     _can_pass_early(actual, expected; kwargs...)
 Used to check if `actual` is basically equal to `expected`, so we don't need to check deeper;
@@ -77,15 +81,28 @@ function check_equal(
     @test ActualPrimal === ExpectedPrimal
 end
 
+
+# Some structual differential and a natural differential
+function check_equal(actual::Composite{P, T}, expected; kwargs...) where {T, P}
+    if _can_pass_early(actual, expected)
+        @test true
+    else
+        @assert (T <: NamedTuple)  # it should be a structual differential if we hit this
+
+        # We are only checking the properties that are in the Composite
+        # the natural differential is allowed to have other properties that we ignore
+        @testset "$P.$ii" for ii in propertynames(actual)
+            check_equal(getproperty(actual, ii), getproperty(expected, ii); kwargs...)
+        end
+    end
+end
+check_equal(x, y::Composite; kwargs...) = check_equal(y, x; args...)
+
 # This catches comparisons of Composites and Tuples/NamedTuple
 # and gives a error messaage complaining about that
 const LegacyZygoteCompTypes = Union{Tuple,NamedTuple}
 check_equal(::C, expected::T) where {C<:Composite,T<:LegacyZygoteCompTypes} = @test C === T
 check_equal(::T, expected::C) where {C<:Composite,T<:LegacyZygoteCompTypes} = @test T === C
-
-check_equal(::Zero, x; kwargs...) = check_equal(zero(x), x; kwargs...)
-check_equal(x, ::Zero; kwargs...) = check_equal(x, zero(x); kwargs...)
-check_equal(x::Zero, y::Zero; kwargs...) = @test true
 
 # Generic fallback, probably a tuple or something
 function check_equal(actual::A, expected::E; kwargs...) where {A, E}
@@ -100,6 +117,7 @@ function check_equal(actual::A, expected::E; kwargs...) where {A, E}
         check_equal(c_actual, c_expected; kwargs...)
     end
 end
+
 
 """
 _check_add!!_behavour(acc, val)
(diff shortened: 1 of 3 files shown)


src/check_result.jl
```diff
@@ -25,6 +25,10 @@ for (T1, T2) in ((AbstractThunk, Any), (AbstractThunk, AbstractThunk), (Any, Abs
     end
 end
 
+check_equal(::Zero, x; kwargs...) = check_equal(zero(x), x; kwargs...)
+check_equal(x, ::Zero; kwargs...) = check_equal(x, zero(x); kwargs...)
+check_equal(x::Zero, y::Zero; kwargs...) = @test true
```
mzgubic, line 30: No real objections to having this, but does it ever occur? I though `check_equal` was for checking AD vs FD results, so not sure when FD would give `Zero()`? Or is it ever used outside this context?

src/check_result.jl
```diff
@@ -76,16 +80,29 @@ function check_equal(
...
+
 # This catches comparisons of Composites and Tuples/NamedTuple
 # and gives a error messaage complaining about that
 const LegacyZygoteCompTypes = Union{Tuple,NamedTuple}
 check_equal(::C, expected::T) where {C<:Composite,T<:LegacyZygoteCompTypes} = @test C === T
 check_equal(::T, expected::C) where {C<:Composite,T<:LegacyZygoteCompTypes} = @test T === C
 
-check_equal(::Zero, x; kwargs...) = check_equal(zero(x), x; kwargs...)
-check_equal(x, ::Zero; kwargs...) = check_equal(x, zero(x); kwargs...)
-check_equal(x::Zero, y::Zero; kwargs...) = @test true
```
mzgubic, line 67: Oh, I realise this was just moved, was that intentional?

src/check_result.jl
```diff
@@ -80,6 +117,7 @@ function check_equal(actual::A, expected::E; kwargs...) where {A, E}
     end
 end
 
+
```
mzgubic, line 120:
```suggestion
```

src/check_result.jl
```diff
@@ -56,19 +80,32 @@ function check_equal(
     @test ActualPrimal === ExpectedPrimal
 end
 
-# This catches comparisons of Composites and Tuples/NamedTuple
-# and gives a error messaage complaining about that
-check_equal(::C, expected::T) where {C<:Composite, T} = @test C === T
-check_equal(::T, expected::C) where {C<:Composite, T} = @test C === T
 
+# Some structual differential and a natural differential
```
mzgubic, line 84: What are structural and natural differentials? Structural means Composite for a struct? Not sure about natural?

src/check_result.jl
```diff
@@ -56,19 +80,32 @@ function check_equal(
...
+    if _can_pass_early(actual, expected)
+        @test true
+    else
+        @assert (T <: NamedTuple)  # it should be a structual differential if we hit this
 
-check_equal(::Zero, x; kwargs...) = check_equal(zero(x), x; kwargs...)
-check_equal(x, ::Zero; kwargs...) = check_equal(x, zero(x); kwargs...)
-check_equal(x::Zero, y::Zero; kwargs...) = @test true
+        # We are only checking the properties that are in the Composite
+        # the natural differential is allowed to have other properties that we ignore
```
mzgubic, line 92: What other properties could it have? Maybe becomes clearer once I know what a natural differential is

src/check_result.jl
```diff
@@ -56,19 +80,32 @@ function check_equal(
...
+        @testset "$P.$ii" for ii in propertynames(actual)
+            check_equal(getproperty(actual, ii), getproperty(expected, ii); kwargs...)
```
mzgubic, line 94:
```suggestion
        @testset "$P.$pn" for pn in propertynames(actual)
            check_equal(getproperty(actual, pn), getproperty(expected, pn); kwargs...)
```

mzgubic (approved):

src/check_result.jl
```diff
@@ -25,6 +25,10 @@ for (T1, T2) in ((AbstractThunk, Any), (AbstractThunk, AbstractThunk), (Any, Abs
     end
 end
 
+check_equal(::Zero, x; kwargs...) = check_equal(zero(x), x; kwargs...)
```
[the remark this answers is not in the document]
oxinabox: This was not changed in this PR, just moved.
Basically, we don't know if we want to taht or not right now.
So leaving it an error so can decide later

src/check_result.jl
oxinabox: yes, this is used in some other places.
it checks primal results,
and it is used to check `add!!(x,y) vs `x+y`
and in that later case exactly this can indeed occur

src/check_result.jl
oxinabox: yah, i realized it was in the middle of all the otehr things that are defined recursively

src/check_result.jl
oxinabox: natural is something a human decided made sense.
Structural just means `composite`
see https://www.juliadiff.org/ChainRulesCore.jl/dev/design/many_differentials.html

src/check_result.jl
oxinabox: All kinds of junk potentially.
Like metadata; or reduncant precomputed values.
But i am also particularly thinking of Factorizations.

src/check_result.jl
```diff
@@ -56,19 +80,32 @@ function check_equal(
...
-check_equal(x, ::Zero; kwargs...) = check_equal(x, zero(x); kwargs...)
-check_equal(x::Zero, y::Zero; kwargs...) = @test true
+        # We are only checking the properties that are in the Composite
+        # the natural differential is allowed to have other properties that we ignore
+        @testset "$P.$pn" for pn in propertynames(actual)
+            check_equal(getproperty(actual, pn), getproperty(expected, pn); kwargs...)
+        end
+    end
+end
+check_equal(x, y::Composite; kwargs...) = check_equal(y, x; args...)
```
oxinabox, line 98:
```suggestion
check_equal(x, y::Composite; kwargs...) = check_equal(y, x; kwargs...)
```

src/check_result.jl
```diff
@@ -77,16 +81,29 @@ function check_equal(
     @test ActualPrimal === ExpectedPrimal
 end
 
+
+# Some structual differential and a natural differential
+function check_equal(actual::Composite{P, T}, expected; kwargs...) where {T, P}
+    if _can_pass_early(actual, expected)
+        @test true
+    else
+        @assert (T <: NamedTuple)  # it should be a structual differential if we hit this
```
mzgubic, line 90: Is there a reason why this couldn't be a Tuple? I think it is causing https://github.com/JuliaDiff/ChainRulesCore.jl/issues/266

PR #86, opened 2020-12-17
Add GitHub Actions CI workflows

sethaxen:
The workflows are based on those from ChainRules.jl. Since we don't build documentation or compute code coverage with Travis, I didn't add those in this PR.


oxinabox:
Looks like fails on Julia-latest would be fixed by
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/72

But also i think we should stop testing on the latest julia

sethaxen:
> Looks like fails on Julia-latest would be fixed by
> #72
> 
> But also i think we should stop testing on the latest julia

Do you mean JuliaNightly? I went ahead and removed that workflow.

oxinabox (approved):
Lets also delete Travis CI in this PR

sethaxen:
Done. Also I just realized that there was a Documenter stage with Travis, but it's been broken since Julia 1.5, so I added a corresponding GitHub actions workflow.

sethaxen:
Weird, Julia 1.0 tests just started failing. @mzgubic @oxinabox is this due to  #87?

oxinabox:
> Weird, Julia 1.0 (Edit: and Julia 1) tests just started failing. @mzgubic @oxinabox is this due to #87?

Looks related.
Let me pull branches and test locally

oxinabox:
yes, I have reproduced that locally, I am not sure the root cause but it seems like it does not block this PR.
what ever it is, seems like we can attribute bad code being merged to CI not running
Lets merge this PR now so it stops from happenning again

PR #87, opened 2020-12-17
logabsdet test fix

mzgubic:
Closes https://github.com/JuliaDiff/ChainRulesCore.jl/issues/266

There were two issues:
1) missing `;kwargs` causing the wrong method to dispatch
2) finite differencing returns a Tuple for functions which output a Tuple

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -97,13 +97,15 @@ function _make_jvp_call(fdm, f, xs, ẋs, ignores)
     ignores = collect(ignores)
     all(ignores) && return ntuple(_->nothing, length(xs))
     sigargs = zip(xs[.!ignores], ẋs[.!ignores])
-    return to_composite(jvp(fdm, f2, sigargs...))
+    return _maybe_fix_to_composite(jvp(fdm, f2, sigargs...))
 end
 
-# remove after https://github.com/JuliaDiff/FiniteDifferences.jl/issues/97
-to_composite(x::Tuple) = Composite{typeof(x)}(x...)
-to_composite(x::NamedTuple) = Composite{typeof(x)}(;x...)
-to_composite(x) = x
+# TODO: remove after https://github.com/JuliaDiff/FiniteDifferences.jl/issues/97
+# For functions which return a tuple, FD returns a tuple to represent the differential. Tuple
+# is not a natural differential, because it doesn't overload +, so make it a Composite.
+_maybe_fix_to_composite(x::Tuple) = Composite{typeof(x)}(x...)
+_maybe_fix_to_composite(x::NamedTuple) = Composite{typeof(x)}(;x...)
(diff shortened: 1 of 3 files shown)


oxinabox:
Needs tests.
We can make a MWE of a rule for `x->(x, x)` where `test_frule` currently errors/fails

mzgubic:
right, good to keep the test inside ChainRulesTestUtils and not rely on integration to flag this.

I needed to do x->(x, 1.0) because it is important for the backing type of the composite to be different from the differential so that it doesn't `_can_pass_early`

src/testers.jl
```diff
@@ -97,9 +97,14 @@ function _make_jvp_call(fdm, f, xs, ẋs, ignores)
     ignores = collect(ignores)
     all(ignores) && return ntuple(_->nothing, length(xs))
     sigargs = zip(xs[.!ignores], ẋs[.!ignores])
-    return jvp(fdm, f2, sigargs...)
+    return to_composite(jvp(fdm, f2, sigargs...))
 end
 
+# remove after https://github.com/JuliaDiff/FiniteDifferences.jl/issues/97
```
oxinabox, line 103: using `#TODO` causes it to get picked up by various automated tooling
```suggestion
#TODO remove after https://github.com/JuliaDiff/FiniteDifferences.jl/issues/97
```

src/testers.jl
```diff
@@ -97,9 +97,14 @@ function _make_jvp_call(fdm, f, xs, ẋs, ignores)
...
+to_composite(x::Tuple) = Composite{typeof(x)}(x...)
+to_composite(x::NamedTuple) = Composite{typeof(x)}(;x...)
+to_composite(x) = x
```
oxinabox, line 106: Since they are not expored
```suggestion
_to_composite(x::Tuple) = Composite{typeof(x)}(x...)
_to_composite(x::NamedTuple) = Composite{typeof(x)}(;x...)
_to_composite(x) = x
```

Might want to say:

```suggestion
_maybe_as_composite(x::Tuple) = Composite{typeof(x)}(x...)
_maybe_as_composite(x::NamedTuple) = Composite{typeof(x)}(;x...)
_maybe_as_composite(x) = x
```
or something?

src/testers.jl
oxinabox: also maybe a comment about this fixing up FiniteDiff sometimes returning Tuples etc rathter than valid diff types as part of that TODO.
to explai what this function does?

src/testers.jl
oxinabox: or maybe including `fix` in the name?

mzgubic: all good suggestions, so added them all

oxinabox (approved):

PR #89, opened 2020-12-17
Fix rrule test on tuples

oxinabox:
[These tests](https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/master/test/testers.jl#L250) were failing even after  #87 

CI being broken seems to have let a bunch of stuff slip through


oxinabox:
This is what the CI does without this fix
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/runs/1572957783

mzgubic (approved):
That's my bad, sorry! I was testing things locally and it looks like I only ran the tests that were failing manually rather than the whole test suite. It's so easy to rely on CI

PR #90, opened 2020-12-18
Send Invenia Slack Notifications if nightly build fails

oxinabox:
This sets it up so an RSE at Invenia will be notified if nightly build fails

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -5,7 +5,7 @@ on:
     tags: [v*]
   pull_request:
   schedule:
-    - cron: "0 0 * * *" # run on default branch every night at midnight UTC
+    - cron: '0 2 * * *'  # Daily at 2 AM UTC (8 PM CST)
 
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -46,3 +46,19 @@ jobs:
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
+
+  slack:
+    name: Notify Slack Failure
+    needs: test
+    runs-on: ubuntu-latest
+    if: github.event == 'schedule'
```
oxinabox, line 54:
```suggestion
    if:  always() && github.event == 'schedule'
```

fchorney, line 54:
```suggestion
    if: always() && github.event_name == 'schedule'
```

fchorney (approved):

PR #91, opened 2020-12-20
Fix docs

matbesancon:
This fixes the issue of absent doc pages, the reason being `docs/make.jl` using Travis-specific code


matbesancon:
Not sure what integration tests do but the others seem fine

matbesancon:
I just added the push preview to fall back to the same behaviour as intended with the previous code (being able to see previews)

oxinabox:
> Not sure what integration tests do but the others seem fine

They are reverse dependency tests on in this case that ChainRules.jl works still with this  version of ChainRuleTestUtils.
Which it doesn't but that's not a problem since it's also marked as not compatible

oxinabox (approved):

PR #93, opened 2020-12-21
Get primal type from primal when fixing composites

oxinabox:
Fixes https://github.com/JuliaMath/SpecialFunctions.jl/issues/288#issuecomment-749091586

> This is being reported now because of https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/84
> which added checking that the primal type when it was checking if two `Composite{Primal}` were equal.
> but looking at it the code for determining the primal type if the Composite came from FiniteDifferencing is wrong
> https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/92c4888361557f010406aaf14a16240431067915/src/testers.jl#L106-L108
> because it just assumes that that type of the difference is the type of the primal.
> Which it is not for the case for Integers.
> 
> That code was added in https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/89 and https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/87 That PR was not very successful, since this the third bug fix i will need to make to it.

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -70,19 +70,20 @@ function _make_j′vp_call(fdm, f, ȳ, xs, ignores)
     @assert length(fd) == length(arginds)
 
     for (dx, ind) in zip(fd, arginds)
-        args[ind] = _maybe_fix_to_composite(dx)
+        args[ind] = _maybe_fix_to_composite(xs[ind], dx)
     end
     return (args...,)
 end
 
 """
-    _make_jvp_call(fdm, f, xs, ẋs, ignores)
+    _make_jvp_call(fdm, f, y, xs, ẋs, ignores)
 
 Call `FiniteDifferences.jvp`, with the option to ignore certain `xs`.
 
 # Arguments
 - `fdm::FiniteDifferenceMethod`: How to numerically differentiate `f`.
 - `f`: The function to differentiate.
+- `y`: The primal output `y=f(xs...)` or at least something of the right type
 - `xs`: Inputs to `f`, such that `y = f(xs...)`.
 - `ẋs`: The directional derivatives of `xs` w.r.t. some real number `t`.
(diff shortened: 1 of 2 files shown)


test/testers.jl
```diff
@@ -241,6 +241,7 @@ end
         end
 
         CTuple{N} = Composite{NTuple{N, Float64}}  # shorter for testing
+        CIntTuple{N} = Composite{NTuple{N, Int}}  # Primal is Int
```
oxinabox, line 244:
```suggestion
```

src/testers.jl
```diff
@@ -91,21 +92,21 @@ Call `FiniteDifferences.jvp`, with the option to ignore certain `xs`.
...
 end
 
 # TODO: remove after https://github.com/JuliaDiff/FiniteDifferences.jl/issues/97
 # For functions which return a tuple, FD returns a tuple to represent the differential. Tuple
 # is not a natural differential, because it doesn't overload +, so make it a Composite.
-_maybe_fix_to_composite(x::Tuple) = Composite{typeof(x)}(x...)
-_maybe_fix_to_composite(x::NamedTuple) = Composite{typeof(x)}(;x...)
-_maybe_fix_to_composite(x) = x
+_maybe_fix_to_composite(::P, x::Tuple) where P = Composite{P}(x...)
+_maybe_fix_to_composite(::P, x::NamedTuple) where P = Composite{P}(;x...)
```
simeonschaub, line 108:
```suggestion
_maybe_fix_to_composite(::P, x::Tuple) where {P} = Composite{P}(x...)
_maybe_fix_to_composite(::P, x::NamedTuple) where {P} = Composite{P}(;x...)
```

simeonschaub (approved):
One small style nitpick, otherwise LGTM.

PR #97, opened 2020-12-26
Improve some errors

### LINKED ISSUE #81: directly report if no rrule/frule is defined
Right now we hit the fallback definition of `rrule`/frule` which is to  just return `nothing`.
which then gives an error complaining about iteration failing.

We should instead detect that an error ourselves with somehting more clear.
Possibly throwing a `MethodError` which would show what is closest.
(though it would also show the `Any` method that does match but maybe that is ok)

mcabbott:
This tries to make the errors from this package just a hair less inscrutable. 

Edit: closes #81

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -200,7 +200,7 @@ end
 If `check_inferred=true`, then the type-stability of the `frule` is checked.
 All remaining keyword arguments are passed to `isapprox`.
 """
-function frule_test(f, xẋs::Tuple{Any, Any}...; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(), check_inferred=true, kwargs...)
+function frule_test(f, xẋs::Tuple{Any, Any}...; rtol::Real=1e-9, atol::Real=1e-9, fdm=_fdm, fkwargs::NamedTuple=NamedTuple(), check_inferred::Bool=true, kwargs...)
     # To simplify some of the calls we make later lets group the kwargs for reuse
     isapprox_kwargs = (; rtol=rtol, atol=atol, kwargs...)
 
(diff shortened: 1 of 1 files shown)


src/testers.jl
```diff
@@ -251,7 +251,9 @@ function rrule_test(f, ȳ, xx̄s::Tuple{Any, Any}...; rtol=1e-9, atol=1e-9, fdm
     xs = first.(xx̄s)
     accumulated_x̄ = last.(xx̄s)
     check_inferred && _test_inferred(rrule, f, xs...; fkwargs...)
-    y_ad, pullback = rrule(f, xs...; fkwargs...)
+    res = rrule(f, xs...; fkwargs...)
+    isnothing(res) && throw("no rrule defined for $f, with arguments $(typeof(xs))")
```
oxinabox, line 255: What do you think if we throw a `MethodError` hear instead.
Just so it gives the list of near candidates?

oxinabox, line 255: Need to use `res===nothing` for 1.0 compatibility

oxinabox (approved):
LGTM except the think with `isnothing`

src/testers.jl
mcabbott: Anything better than `iterate(nothing)` is fine, really.

mcabbott: New changed.

oxinabox:
Oh sorry I missed this before, but shouldn't we also error if rule is not defined for `frule`?

mcabbott:
Probably, I only fixed things which bit me, so far...

mcabbott:
I've added one for frule too now. (And I think I got the signature right!)

BTW I'm not entirely sure MethodError is the right thing here. It has the right information, but isn't quite the truth -- a method does exist, it's just the fallback one. Maybe an explicit description is clearer?

nickrobinson251:
closes https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/81

oxinabox:
> BTW I'm not entirely sure MethodError is the right thing here. It has the right information, but isn't quite the truth -- a method does exist, it's just the fallback one. Maybe an explicit description is clearer?

Manually thrown MethodErrors do that though.
E.g. https://github.com/JuliaMath/SpecialFunctions.jl/blob/master/src/gamma.jl#L528

PR #102, opened 2021-01-11
Increase FiniteDifferences to v0.12

wesselb:
This PR also sets `max_range` of `_fdm` to `1e-2` to prevent tests from accidentally failing due to singularities.

See also https://github.com/JuliaDiff/ChainRules.jl/pull/341.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,5 +13,5 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 [compat]
 ChainRulesCore = "0.9.13"
 Compat = "3"
-FiniteDifferences = "0.11.2"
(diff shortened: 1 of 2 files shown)


oxinabox (approved):
Don't forget to bump the version of ChainRuleTestUtils.jl at the top of the project.
This is nonbreaking to it's API

wesselb:
I think 0.6 is still a DEV version and not released yet, but I've added a compat entry for ChainRulesTestUtils 0.6 in the ChainRules PR to allow the integration test to run. (Do let me know if that's not the right way of doing things.)

oxinabox:
It is fine for the integration test to fail because we haven't made it compatible yet.

Merge this when ever you like

PR #103, opened 2021-01-12
Only check type-stability of frule and rrule if primal is type-stable

sethaxen:
This PR follows up #78 to only test that `frule`s and `rrule`s are type-stable when the primal itself is type-stable.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.6.0-DEV"
+version = "0.6.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -189,6 +189,20 @@ function _test_inferred(f, args...; kwargs...)
 end
 
 """
+    _is_typestable(f, args...; kwargs...) -> Bool
+
+Return whether `f(args...; kwargs...)` is type-stable.
+"""
+function _is_typestable(f, args...; kwargs...)
+    try
+        _test_inferred(f, args...; kwargs...)
+        return true
+    catch ErrorException
+        return false
+    end
+end
+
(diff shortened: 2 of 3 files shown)


sethaxen:
@oxinabox I guess this could be included as part of v0.6.0, but I've incremented the version number anyways.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.6.0-DEV"
+version = "0.6.1"
```
oxinabox, line 3: Can't make a patch release til we do an actual release
but lets do an actual release with this
```suggestion
version = "0.6.0"
```

src/testers.jl
```diff
@@ -188,6 +188,20 @@ function _test_inferred(f, args...; kwargs...)
     end
 end
 
+"""
+    _is_typestable(f, args...; kwargs...) -> Bool
+
+Return whether `f(args...; kwargs...)` is type-stable.
+"""
+function _is_typestable(f, args...; kwargs...)
```
oxinabox, line 196: Technically this is not checking type-stability, it is checking inferability.
Julia doesn't promise that all type-stable things will infer.

src/testers.jl
```diff
@@ -197,7 +211,8 @@ end
 - `ẋ`: differential w.r.t. `x` (should generally be set randomly).
 
 `fkwargs` are passed to `f` as keyword arguments.
-If `check_inferred=true`, then the type-stability of the `frule` is checked.
+If `check_inferred=true`, then the type-stability of the `frule` is checked, as long as `f`
+is itself type-stable.
```
oxinabox, line 215: similar here (I should have caught this before).
Inferable implies type stable
type stable does not imply inferable.

oxinabox (approved):
LGTM baring a minor comment.
Merge once addressed

Project.toml
oxinabox: Actually can keave it at 0.6.1 and I am releasing 0.6 presently

PR #105, opened 2021-01-14
make integration tests hit many packages

### LINKED ISSUE #85: More Reverse Dependency Integration Test
We should have reverse dependency intregration tests on thigns that have a dependency (including test time)
on ChainRulesTestUtils.

Adding more of these should not be hard because it should basically be copying what we already have.
Perhaps smartly using a `matrix`

Us not chatching https://github.com/invenia/BlockDiagonals.jl/issues/49 until nightly ran, when we could have avoided it in #57  is not great.
(Though can't deny that Nightly build is given value there)

oxinabox:
closes #85 

I kind of want to drop a ton of other things from the testing matrix to make more space for these

### DIFF
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches: [master]
     tags: [v*]
-  pull_request
+  pull_request:
 
 jobs:
   test:
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -10,6 +10,7 @@ jobs:
     name: ${{ matrix.package.repo }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         julia-version: [1]
         os: [ubuntu-latest]
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -18,7 +18,6 @@ jobs:
           - {user: JuliaDiff, repo: ChainRules.jl}
           - {user: JuliaMath, repo: SpecialFunctions.jl}
           - {user:  invenia, repo: BlockDiagonals.jl}
-          - {user:  invenia, repo: PDMatsExtras.jl}
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
@@ -44,6 +43,13 @@ jobs:
             # If we can't resolve that means this is incompatible by SemVer and this is fine
             # It means we marked this as a breaking change, so we don't need to worry about
             # Mistakenly introducing a breaking change, as we have intentionally made one
+
+            # Note: this doens't solve all resolver errors, e.g. might be a mutual dependency
+            # of this package and a test only dependency of the downstream packagem that are
+            # incompatible. And this check only catches resolution failures with main dependencies
+            # (plus this package itself) with this package
+
+            @info "Not compatible with this release. No problem." exception=(err, catch_backtrace())
             exit(0)  #  Exit immediately, as a success
           end
           Pkg.test()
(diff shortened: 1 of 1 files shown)


oxinabox:
This seems to be working right.
The integration tests other than ChainRules.jl are all failing because of incompatibility -- which is fine
I wish there was a way to indicate that that is why they are failing, and skip the remainder

oxinabox:
You can see how it removed the case of a ❌  for if the downstream is not compatible with this release
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/105/checks?check_run_id=1710412539#step:6:25

As the commends in the code say this doens't catch every way that an incompatibly could occur.
But it should reduce the "boy who cried wolf" effect a whole lot

.github/workflows/IntegrationTest.yml
```diff
@@ -0,0 +1,55 @@
...
+            # force it to use this PR's version of the package
+            Pkg.develop(PackageSpec(path="."))
+            Pkg.update()
+          catch err
+            err isa Pkg.Resolve.ResolverError || rethrow()
+            # If we can't resolve that means this is incompatible by SemVer and this is fine
+            # It means we marked this as a breaking change, so we don't need to worry about
+            # Mistakenly introducing a breaking change, as we have intentionally made one
+
+            # Note: this doens't solve all resolver errors, e.g. might be a mutual dependency
```
mattBrzezinski, line 47:
```suggestion
            # Note: this doesn't solve all resolver errors, e.g. might be a mutual dependency
```

.github/workflows/IntegrationTest.yml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version: [1]
+        os: [ubuntu-latest]
+        package:
+          - {user: JuliaDiff, repo: ChainRules.jl}
+          - {user: JuliaMath, repo: SpecialFunctions.jl}
+          - {user:  invenia, repo: BlockDiagonals.jl}
```
mattBrzezinski, line 20:
```suggestion
          - {user: invenia, repo: BlockDiagonals.jl}
```

.github/workflows/IntegrationTest.yml
```diff
@@ -0,0 +1,55 @@
+name: IntegrationTest
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
```
mattBrzezinski, line 6:
```suggestion
  pull_request
```

.github/workflows/IntegrationTest.yml
```diff
@@ -0,0 +1,55 @@
...
+
+jobs:
+  test:
+    name: ${{ matrix.package.repo }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version: [1]
```
mattBrzezinski, line 15: Do we also want to test on LTS?

mattBrzezinski (approved):
LGTM

.github/workflows/IntegrationTest.yml
oxinabox: we do not. In part because our dependencies might not support the LTS

.github/workflows/IntegrationTest.yml
```diff
@@ -0,0 +1,55 @@
+name: IntegrationTest
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request
```
oxinabox, line 6:
```suggestion
  pull_request:
```
@mattBrzezinski  the `:` was needed else i get this error
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/actions/runs/488680037

mattBrzezinski: Oh strange!

PR #107, opened 2021-01-18
Add two usage examples to documentation

mzgubic:
There are also two checks added that I thought would improve user experience when making a mistake in writing a custom rule.


src/testers.jl
```diff
@@ -228,6 +228,7 @@ function frule_test(f, xẋs::Tuple{Any, Any}...; rtol::Real=1e-9, atol::Real=1e
     end
     res = frule((NO_FIELDS, deepcopy(ẋs)...), f, deepcopy(xs)...; deepcopy(fkwargs)...)
     res === nothing && throw(MethodError(frule, typeof((f, xs...))))
+    res isa Tuple || error("The frule should return (y, ∂y), not $res.")
```
mzgubic, line 231: this adds a hopefully useful bit of info

src/testers.jl
```diff
@@ -280,6 +281,7 @@ function rrule_test(f, ȳ, xx̄s::Tuple{Any, Any}...; rtol::Real=1e-9, atol::Re
 
     check_inferred && _test_inferred(pullback, ȳ)
     ∂s = pullback(ȳ)
+    ∂s isa Tuple || error("The pullback must return (∂self, ∂args...), not $∂s.")
```
mzgubic, line 284: this is the other message

docs/src/index.md
```diff
@@ -5,11 +5,109 @@
...
+this in mind when testing discontinuous rules for functions like
+[ReLU](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)), which should ideally
+be tested at both `x` being above and below zero.
+Additionally, choosing `ẋ` in an unfortunate way (e.g. as zeros) could hide
+underlying problems with the defined `frule`.
+
+```
+xs = (3.33, -7.77)
+ẋs = (rand(), rand())
+frule_test(two2three, (xs[1], ẋs[1]), (xs[2], ẋs[2]))
```
oxinabox, line 56: Do you think it would be simpler and easier to understand if we use `a, b` rather than `xs` ?
and similar for dotted versions

docs/src/index.md
```diff
@@ -5,11 +5,109 @@
...
+```
+
+### Testing the `rrule`
+
+`rrule_test` takes in the function `f`, sensitivities of the function outputs `ȳ`,
+and tuples `(x, x̄)` for each function argument `x`.
+`x̄` is the accumulated adjoint which should be set randomly.
```
oxinabox, line 63:
```suggestion
`x̄` is the accumulated adjoint which can be set to an arbitary.
```

docs/src/index.md
```diff
@@ -5,11 +5,109 @@
...
+The call will test the `rrule` for function `f` at the point `x`, and similarly to
+`frule` some rules should be tested at multiple points in the domain.
+Choosing `ȳ` in an unfortunate way (e.g. as zeros) could hide underlying problems with
+the `rrule`. 
+```
+xs = (3.33, -7.77)
+ȳs = (rand(), rand(), rand())
+x̄s = (rand(), rand())
```
oxinabox, line 71: similar for my comments about maybe we avoid using `xs` and just use `a` and `b` seperately

docs/src/index.md
```diff
@@ -5,11 +5,109 @@
...
+rrule_test(two2three, ȳs, (xs[1], x̄s[1]), (xs[2], x̄s[2]))
+```
+
+## Scalar example
+
+For functions with a single argument and a single output, such as e.g. `ReLU`,
```
oxinabox, line 77:
```suggestion
For functions with a single argument and a single output, such as e.g. ReLU,
```
Or

```suggestion
For functions with a single argument and a single output, such as e.g. `relu`,
```

docs/src/index.md
```diff
@@ -5,11 +5,109 @@
...
+```
+function ChainRulesCore.rrule(::typeof(relu), x::Real)
+    y = relu(x)
+    dydx = x <= 0 ? zero(x) : one(x)
+    function relu_pullback(Ȳ)
+        return (NO_FIELDS, Ȳ .* dydx)
+    end
+    return y, relu_pullback
+end
+```
```
oxinabox, line 101: can we use `@scalar_rule` ?
`test_scalar` and `@scalar_rule` tend to go together.

docs/src/index.md
```diff
@@ -5,11 +5,109 @@
 
 
 [ChainRulesTestUtils.jl](https://github.com/JuliaDiff/ChainRulesTestUtils.jl) helps you test [`ChainRulesCore.frule`](http://www.juliadiff.org/ChainRulesCore.jl/dev/api.html) and [`ChainRulesCore.rrule`](http://www.juliadiff.org/ChainRulesCore.jl/dev/api.html) methods, when adding rules for your functions in your own packages.
-
 For information about ChainRules, including how to write rules, refer to the general ChainRules Documentation:
 [![](https://img.shields.io/badge/docs-master-blue.svg)](https://JuliaDiff.github.io/ChainRulesCore.jl/dev)
 [![](https://img.shields.io/badge/docs-stable-blue.svg)](https://JuliaDiff.github.io/ChainRulesCore.jl/stable)
 
+## Canonical example
```
oxinabox, line 12: should we make all of these doc tests or examples?
or make them all run and display examples?

docs/src/index.md
```diff
@@ -5,11 +5,109 @@
...
+function ChainRulesCore.rrule(::typeof(two2three), a, b)
+    y = two2three(a, b)
+    function two2three_pullback(Ȳ)
+        return (NO_FIELDS, 2.0*Ȳ[2], 3.0*Ȳ[3])
+    end
+    return y, two2three_pullback
+end
+```
+
+The `test_frule`/`test_rrule` helper function compares the `frule`/`rrule` outputs
```
oxinabox, line 39: Let's link to the API docs page
```suggestion
The [`test_frule`](@ref)/[`test_rrule`](@ref) helper function compares the `frule`/`rrule` outputs
```

docs/src/index.md
```diff
@@ -5,11 +5,109 @@
...
+    y = relu(x)
+    dydx = x <= 0 ? zero(x) : one(x)
+    function relu_pullback(Ȳ)
+        return (NO_FIELDS, Ȳ .* dydx)
+    end
+    return y, relu_pullback
+end
+```
+
+`test_scalar` function is provided to test both the `frule` and the `rrule` with a single
```
oxinabox, line 103:
```suggestion
[`test_scalar`](@ref) function is provided to test both the `frule` and the `rrule` with a single
```

src/testers.jl
oxinabox: LGTM

oxinabox (approved):

docs/src/index.md
mzgubic: nice catch, of course

docs/src/index.md
mzgubic: I know what doctests are. I don't know what "example" and "run and display examples" are, would you mind explaining?

I am a bit reluctant to add doctests because the printouts are pretty long and confusing, e.g.
```
julia> rrule_test(relu, 2.1, (-1.2, .23))
Test Summary:                      |
Don't thunk only non_zero argument | No tests
Test.DefaultTestSet("Don't thunk only non_zero argument", Any[], 0, false)
```

docs/src/index.md
mzgubic: I went with `x1` and `x2` (changing the function definition to be consistent instead) just because the text feels more natural talking about arguments `x` rather than `a`

docs/src/index.md
mzgubic: I've added doctests

Issue #96, opened 2020-12-22
Make naming consistent: rename `test_scalar` or `frule_test`+`rrule_test`


nickrobinson251:
We have `frule_test` for `frule`s, `rrule_test` for `rrule`... but `test_scalar` for `@scalar_rule` created rules.

I think it'd be easier to remember these names if we were consistent on whether  we used the suffix `_test` or the prefix `test_`.

Either (a) rename `test_scalar` -> `scalar_test` (or `scalar_rule_test`)
Or (b) rename `frule_test` -> `test_frule` and `rrule_test` -> `test_rrule`

Renaming is breaking, and that's annoying, but we may want to consider doing it for a v1.0 of this package?

mzgubic:
Going with `test_frule` and `test_frule` would be consistent with the other test suite tools e.g. `@test_throws`, `@test_logs`, etc. where `test` is always first.

nickrobinson251:
also `check_equal` -> `test_approx`

mzgubic:
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/116

nickrobinson251:
> also `check_equal` -> `test_approx`

Was this decided against?

mzgubic:
Oh, I missed that bit completely. I don't think we have reached conclusion on that. Will just reopen this one

nickrobinson251:
ah cool, didn't know if i'd jsut missed that decision. 

yeah, seems easy to miss in this issue - made a new one 

https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/134

thanks <3

PR #111, opened 2021-01-21
add check_equal docs

mzgubic:
a quick paragraph on `check_equal`

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -52,12 +52,9 @@ They can be used for any type and number of inputs and outputs.
 ### Testing the `frule`
 
 [`frule_test`](@ref) takes in the function `f` and tuples `(x, ẋ)` for each function argument `x`.
-The call will test the `frule` for function `f` at the point `x` in the domain. Keep
-this in mind when testing discontinuous rules for functions like
-[ReLU](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)), which should ideally
-be tested at both `x` being above and below zero.
-Additionally, choosing `ẋ` in an unfortunate way (e.g. as zeros) could hide
-underlying problems with the defined `frule`.
+The call will test the `frule` for function `f` at the point `x` in the domain.
+Keep this in mind when testing discontinuous rules for functions like [ReLU](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)), which should ideally be tested at both `x` being above and below zero.
+Additionally, choosing `ẋ` in an unfortunate way (e.g. as zeros) could hide underlying problems with the defined `frule`.
 
 ```jldoctest ex; output = false
 using ChainRulesTestUtils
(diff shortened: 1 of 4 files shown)


docs/src/index.md
```diff
@@ -134,6 +134,27 @@ Test Summary:                    | Pass  Total
...
 ```
 
+## Custom finite differencing
+
+If a package is using a custom finite differencing method of testing the `frule`s and
+`rrule`s, `check_equal` function provides a convenient way of comparing [various types]
+(https://www.juliadiff.org/ChainRulesCore.jl/dev/design/many_differentials.html#Design-Notes:-The-many-to-many-relationship-between-differential-types-and-primal-types.)
+of differentials.
+
+It is effectively `isapprox` but with better error messages because it knows about `thunk`s
```
nickrobinson251, line 144: why isn't it called `check_approx`, and/or why doesn't it extend `isapprox`?

also why is it called `check_*` rather than `test_*`?

I'm wondering if (i) there are good reasons and they should be explained on this page, or (ii) we should consider renaming this whenever we do #96

mzgubic: We are trying to deprecate `isapprox` methods. I think one of the reasons is that we don't want to pirate `isapprox` for tuple arguments, but there may be other. Should we document that, or is it too low level?

I guess we could call it `test_*` as you suggest, maybe just `test_equal`?

nickrobinson251: thanks :) i think a comment on why this is not just `isapprox` is appropriate. 

Seems `test_approx` would be a better name (i'll add it to #96)

docs/src/index.md
```diff
@@ -134,6 +134,29 @@ Test Summary:                    | Pass  Total
...
+and `ChainRules` differential types `Zero()`, `DoesNotExist()`, and `Composite`.
+It is not simply overloading `isapprox` because then comparing `Tuple`s/`NamedTuple`s
```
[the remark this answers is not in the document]
oxinabox: It is basically `(a,b) -> @test isapprox(a, b)`
so it is different to overloading `isapprox` since it doesn't return true or false, it adds a failure or a pass to the current testset.

The reason it isn't overloading `isapprox` is because sometimes we need to really fiddle with the inputs before putting them into the `@test`
and `@test isapprox(a,b)` will show `a` and `b`'s value on a failure.
But actually we want to fiddle anound with `a` and `b` to give them more readable values before checking them.
Such as `unthunk`ing.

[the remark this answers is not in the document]
mzgubic: Could this not be achieved by something like:
```
function Base.ispprox(x::AbstractThunk, y)
    return isapprox(unthunk(x), y)
end
```
It feels that adding `@test` inside is a matter of style?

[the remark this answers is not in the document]
oxinabox: Nope.

because `@test` only looks 1 level deep when displaying output.

```julia
julia> struct Foo end

julia> unfoo(::Foo) = 1
unfoo (generic function with 1 method)

julia> unfoo(x) = x
unfoo (generic function with 2 methods)

julia> Base.isapprox(foo::Foo, b; kwargs...) = isapprox(unfoo(foo), b)

julia> using Test

julia> @test isapprox(Foo(), 2)
Test Failed at REPL[10]:1
  Expression: isapprox(Foo(), 2)
   Evaluated: isapprox(Foo(), 2)
ERROR: There was an error during testing
```

Notice that it displayed `isapprox(Foo(), 2)` for the `Evaluated` line.

What we want to be displayed is more like:
```julia

julia> a = 1
1

julia> @test isapprox(a, 2)
Test Failed at REPL[12]:1
  Expression: isapprox(a, 2)
   Evaluated: isapprox(1, 2)
ERROR: There was an error during testing
````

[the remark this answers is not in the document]
Comment (mzgubic) [reply docs/src/index.md:146]: Thanks, all clear now!

File: src/testers.jl [hunk 3]
```diff
@@ -256,6 +256,7 @@ end
 - `x`: input at which to evaluate `f` (should generally be set to an arbitary point in the domain).
 - `x̄`: currently accumulated adjoint (should generally be set randomly).
 
+Non-differentiable arguments, such as indices, should have `x̄` set as `nothing`.
```
mzgubic, line 259: @nickrobinson251 I've added this line as well as an abridged version of Lyndon's explanation. Would you mind taking another look?

mzgubic: bump

mzgubic: one more bump

nickrobinson251 (approved):

Issue #67, opened 2020-10-09
Don't make user provide tangents


oxinabox:
Since we can mostly generate random tangents correctly based on the primals.
we should let the using opt to not provide them.

I am not sure the API we want for it.

### Option 1
Maybe we keep the same API as is current, but let the user replace the tangent with, `Auto()`.
Then we go and generate tangents automatically for anything marked `Auto()`.
(or `:auto` or `🎩` or something.)

#### Option 2
Or maybe we just let them skip it, and instead do something else:

#### Option 2A
 via passing in a 1-tuple instead of a 2-tuple for `xẋs`,
and by skipping it for output.

#### Optiom 2B
Or maybe we let them skip it by just passing the primal.
Then it it is a 2-Tuple it needs to be primal, and differential.
But then it is ambigious if the primal itself is a `Tuple`.

#### Option 2C
We could change tuple to be a `Pair` insead, so one would write `x => ẋ`
instead of `(x, ẋ)`.
Then `Pair` is ambigious, but that is less of a problem.

#### Option 2D
We could introduce out own type, and overload some other nice operator to construct it.
One option if `<|` which is ANSI and in 1.0 and kind of looks like a tangent.
Julia 1.5 adds `-->` and `<--`.
Or we could go some unicode character.

oxinabox:
related #66 `frule_test` currently doesn't provide one of them and #59 generates one of it automatically

Issue #66, opened 2020-10-09
Make frule_test take a sample output tangent for testing purposes


oxinabox:
`rrule_test` takes in sample sensitivities (tangents) for all inputs,even though rrule. doesn't use them (since it is reverse mode) which gives us something to test accumulation against.

`ffule_test` should do the same.

needful to be more consistent in how we handle testign that: see #59

willtebbutt:
I think you're correct, although for `frule_test` you would want to accept a tangent for the output.

PR #116, opened 2021-02-03
Automatically provide tangents

oxinabox:
Closes https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/67
Does option 2D witth `x ⊢ ẋ`
that is `\vdash` + <kbd>tab</kbd>
We can argue about symbols though.

- [x] input tangent
- [x] should make frule take an output tangent #66
- [x] should make frule and rrule's output tangent be auto-generated  
- [x] Should probably also do renaming from #96 to `test_frule` and `test_rrule`. Which should make deprecations easier
- [x] tests

#### before:

```julia
x = rand(4, 5)
x̄ = rand(4, 5)

ȳ = rand(2, 10)

rrule_test(reshape, ȳ, (x, x̄), ((2, 10), nothing))
rrule_test(reshape, ȳ, (x, x̄), (2, nothing), (10, nothing))
```

### After
No point declaring `x` in advance any-more.
Can just put the `rand` in-place

```julia
test_rrule(reshape, rand(4, 5), (2, 10) ⊢  nothing)
test_rrule(reshape, rand(4, 5), 2  ⊢  nothing, 10 ⊢  nothing)
```

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -122,16 +122,16 @@ relu at -0.5, with cotangent 1.0 |    4      4
 [`frule_test`](@ref) and [`rrule_test`](@ref) allow you to specify the tangents used for testing.
 This is done by passing in `x ⊢ Δx`, where `x` is the primal and `Δx` is the tangent, in the place of the primal inputs.
 If this is not done the tangent will be automatically generated via [`rand_tangent`](@ref).
-A special-case of this is that if you specify it as `x ⊢ nothing` then finite differencing will not be used on that input.
+A special case of this is that if you specify it as `x ⊢ nothing` then finite differencing will not be used on that input.
 Similarly, by setting the `output_tangent` keyword argument, you can specify the tangent for the primal output.
 
 This can be useful when the default provided [`rand_tangent`](@ref) doesn't produce the desired tangent for your type.
 For example the default tangent for an `Int` is `DoesNotExist()`.
-Which is correct e.g. when the Int represents a descrete integer like in indexing.
-But if you are testing something where the `Int` actually is a special-case of real number, then you would want specify the tangent as a `Float64`.
+Which is correct e.g. when the `Int` represents a discrete integer like in indexing.
+But if you are testing something where the `Int` is actually a special case of a real number, then you would want to specify the tangent as a `Float64`.
 
 Care must be taken when manually specifying tangents.
-In particular when specifying the input tangents to [`test_frule`](@ref) and the output tangent to [`test_rrule`](@ref).
+In particular, when specifying the input tangents to [`test_frule`](@ref) and the output tangent to [`test_rrule`](@ref).
 As these tangents are used to seed the derivative computation.
 Inserting inappropriate zeros can thus hide errors.
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -58,7 +58,7 @@ Keep this in mind when testing discontinuous rules for functions like [ReLU](htt
 ```jldoctest ex; output = false
 using ChainRulesTestUtils
 
-test_frule(two2three, 3.33, -7.77))
+test_frule(two2three, 3.33, -7.77)
 # output
 Test Summary:                    | Pass  Total
 Tuple{Float64,Float64,Float64}.1 |    1      1
@@ -75,7 +75,7 @@ Test Passed
 The call will test the `rrule` for function `f` at the point `x`, and similarly to `frule` some rules should be tested at multiple points in the domain.
 
 ```jldoctest ex; output = false
-test_rrule(two2three, 3.33, -7.77))
+test_rrule(two2three, 3.33, -7.77)
 # output
 Test Summary:                      |
 Don't thunk only non_zero argument | No tests
@@ -119,13 +119,13 @@ relu at -0.5, with cotangent 1.0 |    4      4
 ```
 
 ## Specifying Tangents
-[`frule_test`](@ref) and [`rrule_test`](@ref) allow you to specify the tangents used for testing.
+[`test_frule`](@ref) and [`test_rrule`](@ref) allow you to specify the tangents used for testing.
 This is done by passing in `x ⊢ Δx`, where `x` is the primal and `Δx` is the tangent, in the place of the primal inputs.
-If this is not done the tangent will be automatically generated via [`rand_tangent`](@ref).
+If this is not done the tangent will be automatically generated via [`ChainRulesTestUtils.rand_tangent`](@ref).
 A special case of this is that if you specify it as `x ⊢ nothing` then finite differencing will not be used on that input.
 Similarly, by setting the `output_tangent` keyword argument, you can specify the tangent for the primal output.
 
-This can be useful when the default provided [`rand_tangent`](@ref) doesn't produce the desired tangent for your type.
+This can be useful when the default provided [`ChainRulesTestUtils.rand_tangent`](@ref) doesn't produce the desired tangent for your type.
 For example the default tangent for an `Int` is `DoesNotExist()`.
 Which is correct e.g. when the `Int` represents a discrete integer like in indexing.
 But if you are testing something where the `Int` is actually a special case of a real number, then you would want to specify the tangent as a `Float64`.
@@ -165,4 +165,8 @@ which should have passed the test.
 ```@autodocs
 Modules = [ChainRulesTestUtils]
 Private = false
+```
+
+```@docs
+ChainRulesTestUtils.rand_tangent
 ```
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -45,24 +45,20 @@ end
 
 ```
 
-The [`frule_test`](@ref)/[`rrule_test`](@ref) helper function compares the `frule`/`rrule` outputs
+The [`test_frule`](@ref)/[`test_rrule`](@ref) helper function compares the `frule`/`rrule` outputs
 to the gradients obtained by finite differencing.
 They can be used for any type and number of inputs and outputs.
 
 ### Testing the `frule`
 
-[`frule_test`](@ref) takes in the function `f` and tuples `(x, ẋ)` for each function argument `x`.
+[`test_frule`](@ref) takes in the function `f` and the primal input `x`.
 The call will test the `frule` for function `f` at the point `x` in the domain.
 Keep this in mind when testing discontinuous rules for functions like [ReLU](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)), which should ideally be tested at both `x` being above and below zero.
-Additionally, choosing `ẋ` in an unfortunate way (e.g. as zeros) could hide underlying problems with the defined `frule`.
 
 ```jldoctest ex; output = false
 using ChainRulesTestUtils
 
-x1, x2 = (3.33, -7.77)
-ẋ1, ẋ2 = (rand(), rand())
-
-frule_test(two2three, (x1, ẋ1), (x2, ẋ2))
+test_frule(two2three, 3.33, -7.77))
 # output
 Test Summary:                    | Pass  Total
 Tuple{Float64,Float64,Float64}.1 |    1      1
@@ -75,17 +71,11 @@ Test Passed
 
 ### Testing the `rrule`
 
-[`rrule_test`](@ref) takes in the function `f`, sensitivities of the function outputs `ȳ`, and tuples `(x, x̄)` for each function argument `x`.
-`x̄` is the accumulated adjoint which can be set arbitrarily.
+[`test_rrule`](@ref) takes in the function `f`, and primal inputsr `x`.
 The call will test the `rrule` for function `f` at the point `x`, and similarly to `frule` some rules should be tested at multiple points in the domain.
-Choosing `ȳ` in an unfortunate way (e.g. as zeros) could hide underlying problems with the `rrule`. 
+
 ```jldoctest ex; output = false
-x1, x2 = (3.33, -7.77)
-x̄1, x̄2 = (rand(), rand())
-ȳs = (rand(), rand(), rand())
-
-rrule_test(two2three, ȳs, (x1, x̄1), (x2, x̄2))
-
+test_rrule(two2three, 3.33, -7.77))
 # output
 Test Summary:                      |
 Don't thunk only non_zero argument | No tests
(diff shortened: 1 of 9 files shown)


oxinabox:
I have not added tests for this explictly, but some places we still use this because we explictly want to pass special tangents in.
So we basically have real world based tests that this works.
So i think this is good.
but review should comment if they disagree

oxinabox:
I would say that we should wait til ChainRules is updated to use this, before merging.
But it is going to be a big job to update all the tests. Probably multiple PRs.
So I think we should merge this, and then patch it if we find something was missed.

This is nonbreaking as the old way is deprecated.

src/generate_tangent.jl
```diff
@@ -1,3 +1,48 @@
...
+end
+primal(p::PrimalAndTangent) = p.primal
+tangent(p::PrimalAndTangent) = p.tangent
+
+"""
+    primal ⊢ tangent
+
+Infix shorthand method to construct a `PrimalAndTangent`.
+"""
+const ⊢ = PrimalAndTangent
```
willtebbutt, line 28: Could you add instructions on how to get this symbol to this docstring?

src/ChainRulesTestUtils.jl
```diff
@@ -12,7 +12,8 @@ using Test
 const _fdm = central_fdm(5, 1; max_range=1e-2)
 
 export TestIterator
-export check_equal, test_scalar, frule_test, rrule_test, generate_well_conditioned_matrix
+export check_equal, test_scalar, test_frule, test_rrule, generate_well_conditioned_matrix
+export Auto, ⊢
```
oxinabox, line 16: Don't need to export `Auto()` as I made the output tangent into a kwarg, so user doesn't need to specify to do it Auto, they just need to not pass that kwarg

src/testers.jl
```diff
@@ -59,33 +59,47 @@ function test_scalar(f, z; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(),
         Δv = one(Ω) * im
         @testset "$f at $z, with cotangent $Δv" begin
             # check ∂v_∂x and (if z is complex) ∂v_∂y via reverse mode
-            rrule_test(f, Δv, (z, Δx); rule_test_kwargs...)
+            rrule_test(f, Δv, z ⊢ Δx; rule_test_kwargs...)
```
oxinabox, line 62: This should use the new `test_rrule`

src/testers.jl
```diff
@@ -38,15 +38,15 @@ function test_scalar(f, z; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(),
...
+            frule_test(f, z ⊢ Δy; rule_test_kwargs...)
         end
     end
 
     # test jacobian transpose using reverse mode
     Δu = one(Ω)
     @testset "$f at $z, with cotangent $Δu" begin
         # check ∂u_∂x and (if z is complex) ∂u_∂y via reverse mode
-        rrule_test(f, Δu, (z, Δx); rule_test_kwargs...)
+        rrule_test(f, Δu, z ⊢ Δx; rule_test_kwargs...)
```
oxinabox, line 49: This should use new function

src/testers.jl
```diff
@@ -38,15 +38,15 @@ function test_scalar(f, z; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(),
         Δy = one(z) * im
         @testset "$f at $z, with tangent $Δy" begin
             # check ∂u_∂y and (if Ω is complex) ∂v_∂y via forward mode
-            frule_test(f, (z, Δy); rule_test_kwargs...)
+            frule_test(f, z ⊢ Δy; rule_test_kwargs...)
```
oxinabox, line 41: This should use new function

src/testers.jl
```diff
@@ -109,30 +123,44 @@ function frule_test(f, xẋs::Tuple{Any, Any}...; rtol::Real=1e-9, atol::Real=1e
...
+ - `fdm::FiniteDifferenceMethod`: the finite differencing method to use.
+ - If `check_inferred=true`, then the inferrability of the `rrule` is checked
+   — if `f` is itself inferrable — along with the inferrability of the pullback it returns.
+ - `fkwargs` are passed to `f` as keyword arguments.
+ - All remaining keyword arguments are passed to `isapprox`.
 """
-function rrule_test(f, ȳ, xx̄s::Tuple{Any, Any}...; rtol::Real=1e-9, atol::Real=1e-9, fdm=_fdm, check_inferred::Bool=true, fkwargs::NamedTuple=NamedTuple(), kwargs...)
+function test_rrule(
+    f, inputs...;
+    output_tangent=Auto(),
```
oxinabox, line 149: Should this be `output_cotangent`?

src/testers.jl
```diff
@@ -142,6 +170,8 @@ function rrule_test(f, ȳ, xx̄s::Tuple{Any, Any}...; rtol::Real=1e-9, atol::Re
     y = f(xs...; fkwargs...)
     check_equal(y_ad, y; isapprox_kwargs...)  # make sure primal is correct
 
+    ȳ = tangent(auto_primal_and_tangent(y ⊢ output_tangent))
```
oxinabox, line 173: It's a bit stilly to be attaching a tangent to the primal explicitly, then pealing it off, but it makes both given and  `Auto()` work

oxinabox:
Ah, I forgot to push some changes.
Will do so tomorrow

mzgubic:
That's wonderful! Please ping me when it is done, happy to review

src/ChainRulesTestUtils.jl
oxinabox:
```suggestion
export ⊢
```

src/testers.jl
```diff
@@ -59,33 +59,47 @@ function test_scalar(f, z; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(),
...
+    output_tangent=Auto(),
+    fdm=_fdm,
+    check_inferred::Bool=true,
+    fkwargs::NamedTuple=NamedTuple(),
+    rtol::Real=1e-9, atol::Real=1e-9, kwargs...
+)
     # To simplify some of the calls we make later lets group the kwargs for reuse
     isapprox_kwargs = (; rtol=rtol, atol=atol, kwargs...)
 
     _ensure_not_running_on_functor(f, "frule_test")
```
oxinabox, line 98:
```suggestion
    _ensure_not_running_on_functor(f, "test_frule")
```

src/testers.jl
```diff
@@ -109,30 +123,44 @@ function frule_test(f, xẋs::Tuple{Any, Any}...; rtol::Real=1e-9, atol::Real=1e
...
+    output_tangent=Auto(),
+    fdm=_fdm,
+    check_inferred::Bool=true,
+    fkwargs::NamedTuple=NamedTuple(),
+    rtol::Real=1e-9, atol::Real=1e-9, kwargs...
+)
     # To simplify some of the calls we make later lets group the kwargs for reuse
     isapprox_kwargs = (; rtol=rtol, atol=atol, kwargs...)
 
     _ensure_not_running_on_functor(f, "rrule_test")
```
oxinabox, line 158:
```suggestion
    _ensure_not_running_on_functor(f, "test_rrule")
```

test/deprecated.jl
```diff
@@ -22,3 +22,196 @@
         end
     end
 end
+
+
+
```
oxinabox, line 27:
```suggestion
```

test/deprecated.jl
```diff
@@ -22,3 +22,196 @@
...
+@testset "old testers.jl" begin
```
oxinabox, line 26: This is a subset of the content that was in `testers.jl` using the old API

mzgubic (commented):
I love this change, makes the API so much easier to use. Should we also update the examples in the `docs` to reflect the new usage?

src/ChainRulesTestUtils.jl
```diff
@@ -12,7 +12,8 @@ using Test
 const _fdm = central_fdm(5, 1; max_range=1e-2)
 
 export TestIterator
-export check_equal, test_scalar, frule_test, rrule_test, generate_well_conditioned_matrix
+export check_equal, test_scalar, test_frule, test_rrule, generate_well_conditioned_matrix
```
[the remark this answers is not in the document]
oxinabox: Yes `@deprecate` exports unless you pass the optional 3rd argument as `false`
which is handy as it means you only need to delete it in once place

src/deprecated.jl
```diff
@@ -27,3 +27,15 @@ end
 
 # Must be for same primal
 Base.isapprox(d_ad::Composite{P}, d_fd::Composite{Q}; kwargs...) where {P, Q} = false
+
+
+# From when primal and tangent was passed as a tuple
+@deprecate(
+    rrule_test(f, ȳ, inputs::Tuple{Any,Any}...; kwargs...),
```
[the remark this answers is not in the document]
oxinabox: to see deprecation warnings in Julia 1.5+ you need to start julia with `julia --depwarn=yes`.
(By default `]test` sets this flag.)

src/testers.jl
```diff
@@ -59,34 +59,48 @@ function test_scalar(f, z; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(),
...
-            rrule_test(f, Δv, (z, Δx); rule_test_kwargs...)
+            test_rrule(f, z ⊢ Δx; output_tangent=Δv,  rule_test_kwargs...)
         end
     end
 end
 
 
 """
-    frule_test(f, (x, ẋ)...; rtol=1e-9, atol=1e-9, fdm=central_fdm(5, 1), fkwargs=NamedTuple(), check_inferred=true, kwargs...)
+    test_frule(f, inputs...; output_tangent=Auto(), fdm=central_fdm(5, 1), check_inferred=true, fkwargs::NamedTuple=NamedTuple(), rtol::Real=1e-9, atol::Real=1e-9, kwargs...)
```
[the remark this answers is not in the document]
oxinabox: They normally are, I think.
but this one was already over-width; so I i just matched what was there
I am tempted to say we should just delete the listing of keyward arguments and just let them be listed in the text below.

src/testers.jl
```diff
@@ -59,34 +59,48 @@ function test_scalar(f, z; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(),
...
+   - If `check_inferred=true`, then the inferrability of the `rrule` is checked
+   - If `check_inferred=true`, then the inferrability of the `frule` is checked,
+     as long as `f` is itself inferrable.
+   - `fkwargs` are passed to `f` as keyword arguments.
+   - All remaining keyword arguments are passed to `isapprox`.
 """
-function frule_test(f, xẋs::Tuple{Any, Any}...; rtol::Real=1e-9, atol::Real=1e-9, fdm=_fdm, fkwargs::NamedTuple=NamedTuple(), check_inferred::Bool=true, kwargs...)
+function test_frule(
+    f, inputs...;
+    output_tangent=Auto(),
```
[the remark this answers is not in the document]
oxinabox: we do for purposes of testing `InplaceThunks`
It is issue #66

src/testers.jl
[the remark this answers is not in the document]
mzgubic: thanks, sounds good

test/deprecated.jl
oxinabox: yes, just moved and subset,
no changes

src/testers.jl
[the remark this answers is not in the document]
mzgubic: It doesn't seem to be used in the body of the function for now? Is this meant to be added in a future PR?

src/testers.jl
oxinabox: the alternative is:
```suggestion
    ȳ = output_tangent === Auto() ? rand_tangent(primal) : output_tangent
```

Which maybe is cleaner?

idk yesterday I was hating on branches.
I started to write `auto_tangent(primal, output_tangent)` and realized that it overlapped with `auto_primal_and_tangent`

src/testers.jl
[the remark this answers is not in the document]
oxinabox: It should be, let me check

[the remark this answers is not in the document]
oxinabox: It isn't.
Adding to https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/116/files/3644e23d3bf5ba775bb432a0abe808898e7b2834#diff-988b853ad3c22aebac9421d3a31e60f9330789df43bccfb9fb75365048398fbaR121

docs/src/index.md
```diff
@@ -128,13 +118,30 @@ Test Summary:                    | Pass  Total
 relu at -0.5, with cotangent 1.0 |    4      4
 ```
 
+## Specifying Tangents
+[`frule_test`](@ref) and [`rrule_test`](@ref) allow you to specify the tangents used for testing.
+This is done by passing in `x ⊢ Δx`, where `x` is the primal and `Δx` is the tangent, in the place of the primal inputs.
+If this is not done the tangent will be automatically generated via [`rand_tangent`](@ref).
+A special-case of this is that if you specify it as `x ⊢ nothing` then finite differencing will not be used on that input.
```
mzgubic, line 125:
```suggestion
A special case of this is that if you specify it as `x ⊢ nothing` then finite differencing will not be used on that input.
```

docs/src/index.md
```diff
@@ -128,13 +118,30 @@ Test Summary:                    | Pass  Total
...
+Similarly, by setting the `output_tangent` keyword argument, you can specify the tangent for the primal output.
+
+This can be useful when the default provided [`rand_tangent`](@ref) doesn't produce the desired tangent for your type.
+For example the default tangent for an `Int` is `DoesNotExist()`.
+Which is correct e.g. when the Int represents a descrete integer like in indexing.
```
mzgubic, line 130:
```suggestion
Which is correct e.g. when the `Int` represents a discrete integer like in indexing.
```

docs/src/index.md
```diff
@@ -128,13 +118,30 @@ Test Summary:                    | Pass  Total
...
+But if you are testing something where the `Int` actually is a special-case of real number, then you would want specify the tangent as a `Float64`.
```
mzgubic, line 131:
```suggestion
But if you are testing something where the `Int` is actually a special case of a real number, then you would want to specify the tangent as a `Float64`.
```

docs/src/index.md
```diff
@@ -128,13 +118,30 @@ Test Summary:                    | Pass  Total
...
+
+Care must be taken when manually specifying tangents.
+In particular when specifying the input tangents to [`test_frule`](@ref) and the output tangent to [`test_rrule`](@ref).
```
mzgubic, line 134:
```suggestion
In particular, when specifying the input tangents to [`test_frule`](@ref) and the output tangent to [`test_rrule`](@ref).
```

mzgubic (approved):

src/testers.jl
```diff
@@ -102,38 +116,48 @@ function frule_test(f, xẋs::Tuple{Any, Any}...; rtol::Real=1e-9, atol::Real=1e
     dΩ_fd = _make_jvp_call(fdm, (xs...) -> f(deepcopy(xs)...; deepcopy(fkwargs)...), Ω, xs, ẋs, ẋs_is_ignored)
     check_equal(dΩ_ad, dΩ_fd; isapprox_kwargs...)
 
-    # No tangent is passed in to test accumlation, so generate one
-    # See: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/66
-    acc = rand_tangent(Ω)
+    acc = tangent(auto_primal_and_tangent(Ω ⊢ output_tangent))
```
oxinabox, line 119: @mzgubic fixed

mzgubic: yeah thanks, saw it - lgtm

src/testers.jl
nickrobinson251: `ȳ = output_tangent === Auto() ? rand_tangent(primal) : output_tangent` is certainly much clearer, to my eyes

CarloLucibello:
I just got the deprecation message
```
┌ Warning: `rrule_test(f, ȳ, inputs::Tuple{Any, Any}...; kwargs...)` is deprecated, use `test_rrule(f, (x ⊢ dx for (x, dx) = inputs)...; output_tangent = ȳ, kwargs...)` instead.
```
and it looked a bit too exotic. Why not pairs `=>` instead of vdash?

src/testers.jl
```diff
@@ -102,38 +116,48 @@ function frule_test(f, xẋs::Tuple{Any, Any}...; rtol::Real=1e-9, atol::Real=1e
...
-  Should be same structure as `f(x)` (so if multiple returns should be a tuple)
-- `x`: input at which to evaluate `f` (should generally be set to an arbitary point in the domain).
-- `x̄`: currently accumulated adjoint (should generally be set randomly).
-
-Non-differentiable arguments, such as indices, should have `x̄` set as `nothing`.
-`fkwargs` are passed to `f` as keyword arguments.
-If `check_inferred=true`, then the inferrability of the `rrule` is checked — if `f` is
-itself inferrable — along with the inferrability of the pullback it returns.
-All remaining keyword arguments are passed to `isapprox`.
+- `inputs` either the primal inputs `x`, or primals and their tangents: `x ⊢ ẋ`
```
CarloLucibello, line 130: ẋ should be x̄. 
Maybe it should be explained also here how to produce a vdash

oxinabox:
@CarloLucibello note the deprecations warning is misleading as to the correct action.
Correct is to just delete the partials and let the automatic process kick in.

nickrobinson251:
> @CarloLucibello note the deprecations warning is misleading as to the correct action.
> Correct is to just delete the partials and let the automatic process kick in.

Maybe manually define the deprecation, with a `depwarn` giving the better instructions, rather than rely on `@deprecate`

Otherwise people will do what the deprecation warning tells them to

oxinabox:
Good idea, I will see if I find time

PR #118, opened 2021-02-08
simplify depwarn

mzgubic:
Do we want to have this show up instead?

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -32,10 +32,10 @@ Base.isapprox(d_ad::Composite{P}, d_fd::Composite{Q}; kwargs...) where {P, Q} = 
 # From when primal and tangent was passed as a tuple
 @deprecate(
     rrule_test(f, ȳ, inputs::Tuple{Any,Any}...; kwargs...),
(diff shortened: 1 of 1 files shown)


nickrobinson251:
this probably isn't what we want, since it'll sometimes be incorrect, we probably want to manually give a helpful message saying what you should do _and_ what is exactly equivalent

```julia
function rrule_test(f, ȳ, inputs::Tuple{Any,Any}...; kwargs...)
    msg = """
        `rrule_test(f, ȳ, inputs::Tuple{Any,Any}...; kwargs...)` is deprecated.
        You no longer need to pass in tangents, and can most likely update your code to
        `rrule_test(f, first.(inputs)...; kwargs...)`
        if you do still need to pass in tangents, you can do so like
        `test_rrule(f, ((x ⊢ dx) for (x, dx) in inputs)...; output_tangent=ȳ, kwargs...)`
        """
    Base.depwarn(msg, :rrule_test)
    return test_rrule(f, ((x ⊢ dx) for (x, dx) in inputs)...; output_tangent=ȳ, kwargs...)
```

(idk if this is exactly the right advice)

oxinabox (changes requested):
@nickrobinson251  has it right

Notice how this broke almost all the tests and integration tests.

PR #122, opened 2021-03-08
wrap testers in top level testset

### LINKED ISSUE #121: test_rrule says that it didn't run any tests
Using the example from the docs:
```
using ChainRulesCore
using ChainRulesTestUtils

function two2three(x1::Float64, x2::Float64)
    return 1.0, 2.0*x1, 3.0*x2
end

function ChainRulesCore.rrule(::typeof(two2three), x1, x2)
    y = two2three(x1, x2)
    function two2three_pullback(Ȳ)
        return (NO_FIELDS, 2.0*Ȳ[2], 3.0*Ȳ[3])
    end
    return y, two2three_pullback
end

test_rrule(two2three, 3.33, -7.77)
```
```
Test Summary:                      |
Don't thunk only non_zero argument | No tests
Test.DefaultTestSet("Don't thunk only non_zero argument", Any[], 0, false)
```
version info:
```
  [d360d2e6] ChainRulesCore v0.9.29
  [cdddcdb0] ChainRulesTestUtils v0.6.3
Julia Version 1.5.2
Commit 539f3ce943 (2020-09-23 23:17 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i7-9800X CPU @ 3.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake-avx512)
```
If this is not implemented yet, then probably the `ChainRules` docs shouldn't point to this package at this stage.

oxinabox:
example output
```julia
julia> test_rrule(two2three, 3.33, -7.77)
Test Summary:                          | Pass  Total
test_rrule: two2three at (3.33, -7.77) |    6      6
Test.DefaultTestSet("test_rrule: two2three at (3.33, -7.77)", Any[Test.DefaultTestSet("Don't thunk only non_zero argument", Any[], 0, false)], 6, false)
````

Closes #121 and #108 

I recomment turning on "hide whitespace changes" when reviewing,
since this indented a ton of code without changing it

---
image

### DIFF
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -5,7 +5,8 @@
 """
     NonPassingTestset(desc, results) <: AbstractTestset
 
-A custom testset that doesn't let anything propagate up to the parent testset (or top-level fallback to throwning an error).
+A custom testset that doesn't let anything propagate up to the parent testset 
+(or to the top-level fallback testset, which throws an error on any non-passing result).
 Not passes, not failures, not even errors.
 
 This is useful for being able to observe the testsets results programatically.
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -3,7 +3,7 @@
 # MetaTesting.jl
 
 """
-    EncasedTestset(desc, results) <: AbstractTestset
+    EncasedTestSet(desc, results) <: AbstractTestset
 
 A custom testset that encases all test results within, not letting them out.
 It doesn't let anything propagate up to the parent testset
@@ -14,20 +14,20 @@ Not passes, not failures, not even errors.
 This is useful for being able to observe the testsets results programatically;
 without them triggering actual passes/failures/errors.
 """
-struct EncasedTestset <: Test.AbstractTestSet
+struct EncasedTestSet <: Test.AbstractTestSet
     description::String
     results::Vector{Any}
 end
-EncasedTestset(desc) = EncasedTestset(desc, [])
+EncasedTestSet(desc) = EncasedTestSet(desc, [])
 
-Test.record(ts::EncasedTestset, t) = (push!(ts.results, t); t)
+Test.record(ts::EncasedTestSet, t) = (push!(ts.results, t); t)
 
-function Test.finish(ts::EncasedTestset)
+function Test.finish(ts::EncasedTestSet)
     if Test.get_testset_depth() != 0
         # Attach this test set to the parent test set *if* it is also a NonPassingTestset
         # Otherwise don't as we don't want to push the errors and failures further up.
         parent_ts = Test.get_testset()
-        parent_ts isa EncasedTestset && Test.record(parent_ts, ts)
+        parent_ts isa EncasedTestSet && Test.record(parent_ts, ts)
         return ts
     end
     return ts
@@ -43,7 +43,7 @@ current testset, and will return a collection of all nonpassing test results.
 """
 function nonpassing_results(f)
     # Specify testset type to hijack system
-    ts = @testset EncasedTestset "nonpassing internal" begin
+    ts = @testset EncasedTestSet "nonpassing internal" begin
         f()
     end
     return _extract_nonpasses(ts)
@@ -52,7 +52,7 @@ end
 "extracts as flat collection of failures from a (potential nested) testset"
 _extract_nonpasses(x::Test.Result) = [x,]
 _extract_nonpasses(x::Test.Pass) = Test.Result[]
-_extract_nonpasses(ts::EncasedTestset) = _extract_nonpasses(ts.results)
+_extract_nonpasses(ts::EncasedTestSet) = _extract_nonpasses(ts.results)
 function _extract_nonpasses(xs::Vector)
     if isempty(xs)
         return Test.Result[]
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -17,7 +17,6 @@ end
 NonPassingTestset(desc) = NonPassingTestset(desc, [])
 
 # Records nothing, and throws an error immediately whenever a Fail or
-# Error occurs. Takes no action in the event of a Pass or Broken result
 Test.record(ts::NonPassingTestset, t) = (push!(ts.results, t); t)
 
 function Test.finish(ts::NonPassingTestset)
--- a/test/meta_testing_tools.jl
+++ b/test/meta_testing_tools.jl
@@ -81,7 +81,7 @@ end
 """
     errors(f, msg_pattern="")
 
-`errors(f, msg_pattern)` returns true if at least 1 error is recorded into a testset,
+Returns true if at least 1 error is recorded into a testset
     with a failure matching the given pattern.
 
 `f` should be a function that takes no argument, and calls some code that uses `@testset`.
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by oxinabox** [comment]
oh boy this breaks a lot of our unhappy path tests

**Comment by oxinabox** [comment]
Ok, most of this PR is now improving out meta-testing ability

File: test/meta_testing_tools.jl [hunk 1]
```diff
@@ -71,14 +65,39 @@ function fails(f)
...
             # Sadly we can't throw the original exception as it is only stored as a String
             error("Error occurred during `fails`")
         end
     end
     return did_fail
 end
 
+"""
+    errors(f, msg_pattern="")
+
```
oxinabox, line 77:
```suggestion
```

test/meta_testing_tools.jl
```diff
@@ -2,8 +2,27 @@
 # if they were less nasty in implementation we might consider moving them to a package
 # MetaTesting.jl
 
-# need to bring this into scope explictly so can use in @testset nonpassing_results
-using Test: DefaultTestSet
+struct NonPassingTestset <: Test.AbstractTestSet
```
oxinabox, line 5: is there a better name for this?
`EnclosedTestset`?

oxinabox, line 5:
```suggestion
"""
    NonPassingTestset(desc, results) <: AbstractTestset

A custom testset that doesn't let anything propagate up to the parent testset (or top-level fallback to throwning an error).
Not passes, not failures, not even errors.

This is useful for being able to observe the testsets results programatically.
"""
struct NonPassingTestset <: Test.AbstractTestSet
```

test/meta_testing_tools.jl
```diff
@@ -2,53 +2,55 @@
...
-        # passing testset in its parent testset. We could remove that if we wanted
-        return nonpasses
+struct NonPassingTestset <: Test.AbstractTestSet
+    description::String
+    results::Vector{Any}
+end
+NonPassingTestset(desc) = NonPassingTestset(desc, [])
+
+# Records nothing, and throws an error immediately whenever a Fail or
+# Error occurs. Takes no action in the event of a Pass or Broken result
```
oxinabox, line 20:
```suggestion
```

oxinabox, line 20:
```suggestion
```

test/meta_testing_tools.jl
```diff
@@ -2,53 +2,54 @@
...
-        # Note: we allow the "nonpassing internal" testset to still be pushed as an empty
-        # passing testset in its parent testset. We could remove that if we wanted
-        return nonpasses
+struct NonPassingTestset <: Test.AbstractTestSet
+    description::String
+    results::Vector{Any}
+end
+NonPassingTestset(desc) = NonPassingTestset(desc, [])
+
+# Records nothing, and throws an error immediately whenever a Fail or
```
oxinabox, line 19:
```suggestion
```

test/meta_testing_tools.jl
```diff
@@ -71,14 +71,38 @@ function fails(f)
...
+    results = nonpassing_results(f)
+
+    for result in results
+        result isa Test.Fail && error("Test actually failed (nor errored): \n $result")
+        result isa Test.Error && occursin(msg_pattern, result.value) && return true
+    end
+    return false  # no matching error occured
+end
+
+
```
nickrobinson251, line 103: maybe overload `Base.show` to not print the `results`? 

```suggestion
Base.show(::IO, ts::NonPassingTestset) = print(typeof(ts), "(\"", ts.description, "\")")
```

I know it's meant for programmatic useage, but it leads to slightly confusing REPL output (stemming from the default pretty-printing of `Test.Result`s)

e.g. currently

```julia
julia> ts = @testset NonPassingTestset "foo" begin
           @test false
           @test true
           @test error()
       end
NonPassingTestset("foo", Any[Test Failed at REPL[23]:2
  Expression: false, Test Passed, Error During Test at REPL[23]:4
  Test threw exception
  Expression: error()

  Stacktrace:
   [1] error() at ./error.jl:42
   [2] top-level scope at REPL[23]:4
   [3] top-level scope at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
   [4] top-level scope at REPL[23]:2
  ])
```
whereas it might be nicer to just have
```julia
julia> ts = @testset NonPassingTestset "foo" begin
           @test false
           @test true
           @test error()
       end
NonPassingTestset("foo")
```

nickrobinson251 (approved):
LGTM

As far as i understand this implements a AbstractTestSet as intended, and this seems a nice useage

test/meta_testing_tools.jl
oxinabox: While that output is confusing, it was useful to me during debugging.
Helped me spot a mistake i made; becuase i wasn't pusing contents up correctly.
Your suggested output would hide that info,
WHile there is a more useful output that is cleaner, working it out would be effort.

test/meta_testing_tools.jl
```diff
@@ -71,14 +71,38 @@ function fails(f)
...
+    errors(f, msg_pattern="")
+
+`errors(f, msg_pattern)` returns true if at least 1 error is recorded into a testset,
+    with a failure matching the given pattern.
+
+`f` should be a function that takes no argument, and calls some code that uses `@testset`.
+`msg_pattern` is a regex or a string, that should be contained in the error message.
+If nothing is passed then it default to the empty string, which matches any error message.
+
+If a test fails (rather than passing or erroring) then `errors` will throw an error.
```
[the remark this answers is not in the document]
oxinabox: my notion is that when you are using this you are trying to **exercise 1 particular unhappy path** that ends in an error.
So if you are hitting a failure instead then that is a mistake.

test/meta_testing_tools.jl
oxinabox: Encased is good

oxinabox:
> As far as i understand this implements a AbstractTestSet as intended, and this seems a nice useage

Yeah, i think this is much nicer than before. And I couldn'tt workout how to handle errors the otherway.
A benifit i didn't realize of doing it this way is that we don't need to `mute` anymore.
Because the `println`s I was hiding were all actually triggered by `record`

test/meta_testing_tools.jl
nickrobinson251: yah, i think a better output would be something along the lines of `NonPassingTestset("foo", Any[Fail, Pass, Error])` but i'm not sure the code is worth it.

test/meta_testing_tools.jl
```diff
@@ -2,53 +2,57 @@
...
-            ts = Test.get_testset()  # this is the current testset "nonpassing internal"
-            nonpasses = _extract_nonpasses(ts)
-            # Prevent the failure being recorded in parent testset.
-            empty!(ts.results)
-            ts.anynonpass = false
-        end
-        # Note: we allow the "nonpassing internal" testset to still be pushed as an empty
-        # passing testset in its parent testset. We could remove that if we wanted
-        return nonpasses
+struct EncasedTestset <: Test.AbstractTestSet
```
nickrobinson251, line 17: Naming convention seems to be `TestSet`
```suggestion
struct EncasedTestSet <: Test.AbstractTestSet
```
(maybe worth being conistent, in case we move this to its own package in future)

anyway, nitpick aside, i think this is good to go!

PR #126, opened 2021-03-14
fix doc

ho-oto:


willtebbutt (approved):
Thanks for catching this!

willtebbutt (changes requested):
Actually, could you please bump the patch. Since this is a change to `src`, we'll issue a patch release.

ho-oto:
done!

mzgubic (approved):

PR #129, opened 2021-03-24
no more `⊢` for types for which rand_tangent gives DoesNotExist

### LINKED ISSUE #113: Remove the use of `nothing` to designate non-differentiability in `rrule_test`
Reusing similar example as in #112:
```julia
function partly_dne(a, i)
    return a[i]
end

function ChainRulesCore.rrule(::typeof(partly_dne), a, i)
    y = partly_dne(a, i)
    function partly_dne_pullback(ȳ)
        grad = zeros(size(a))
        grad[i] = ȳ
        return (NO_FIELDS, grad, DoesNotExist())
    end
    return y, partly_dne_pullback
end
```
It seems un `ChainRules` like that `nothing` should be used instead of `DoesNotExist()`. Should we change this for v1?
```
julia> rrule_test(partly_dne, rand(), (rand(4), rand(4)), (1, nothing))

### LINKED ISSUE #128: test_rrule fails for SVD getproperty
```julia
julia> X = rand(4, 5);

julia> F = svd(X);

julia> test_rrule(getproperty, F, :U ⊢ nothing; check_inferred=false); #works

julia> test_rrule(getproperty, F, :U; check_inferred=false)
test_rrule: getproperty at (SVD{Float64,Float64,Array{Float64,2}}([-0.6192242029556081 0.12479513218307592 0.774222559169189 0.03958522869597275; -0.3817006812210394 0.8074445221197816 -0.42842645034494664 -0.1370719165218814; -0.5778635862189617 -0.5431473872113838 -0.3491042850706373 -0.49919013374958354; -0.370053580109079 -0.19352114728787062 -0.3084743160518571 0.8546657298249292], [2.433363114110269, 0.9107609518410127, 0.502268703721448, 0.3805364703630988], [-0.40783626360515335 -0.4682869755948608 -0.4999760961849655 -0.4438514902786392 -0.4091413550325602; -0.15368262527462623 -0.31244046812219173 -0.5049833781003089 0.42761600232070324 0.6640022191001661; 0.00395187552841697 -0.21586336221566868 0.3285878547255053 -0.6978679185007747 0.5986633290815807; -0.7046162161062675 0.6754218630422322 -0.0880547109902377 -0.11535458879628363 0.16205240399547505]), :U): Error During Test at /Users/mzgubic/JuliaEnvs/test/alex/dev/ChainRules/dev/ChainRulesTestUtils/src/testers.jl:159
  Got exception outside of a @test
  MethodError: no method matching length(::DoesNotExist)
  Closest candidates are:
    length(!Matched::CompositeException) at task.jl:41
    length(!Matched::ExponentialBackOff) at error.jl:259
    length(!Matched::Distributed.WorkerPool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Distributed/src/workerpool.jl:136
    ...
  Stacktrace:
   [1] _similar_for(::UnitRange{Int64}, ::Type{Any}, ::DoesNotExist, ::Base.HasLength) at ./array.jl:597
   [2] _collect(::UnitRange{Int64}, ::DoesNotExist, ::Base.HasEltype, ::Base.HasLength) at ./array.jl:630
   [3] collect(::DoesNotExist) at ./array.jl:624
   [4] check_equal(::DoesNotExist, ::Symbol; kwargs::Base.Iterators.Pairs{Symbol,Float64,Tuple{Symbol,Symbol},NamedTuple{(:rtol, :atol),Tuple{Float64,Float64}}}) at /Users/mzgubic/JuliaEnvs/test/alex/dev/ChainRules/dev/ChainRulesTestUtils/src/check_result.jl:109
   [5] macro expansion at /Users/mzgubic/JuliaEnvs/test/alex/dev/ChainRules/dev/ChainRulesTestUtils/src/testers.jl:199 [inlined]
   [6] macro expansion at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115 [inlined]
   [7] test_rrule(::typeof(getproperty), ::SVD{Float64,Float64,Array{Float64,2}}, ::Vararg{Any,N} where N; output_tangent::ChainRulesTestUtils.Auto, fdm::FiniteDifferences.AdaptedFiniteDifferenceMethod{5,1,FiniteDifferences.UnadaptedFiniteDifferenceMethod{7,5}}, check_inferred::Bool, fkwargs::NamedTuple{(),Tuple{}}, rtol::Float64, atol::Float64, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/mzgubic/JuliaEnvs/test/alex/dev/ChainRules/dev/ChainRulesTestUtils/src/testers.jl:160
   [8] top-level scope at REPL[13]:1
   [9] eval(::Module, ::Any) at ./boot.jl:331
   [10] eval_user_input(::Any, ::REPL.REPLBackend) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:134
   [11] repl_backend_loop(::REPL.REPLBackend) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:195
   [12] start_repl_backend(::REPL.REPLBackend, ::Any) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:180
   [13] run_repl(::REPL.AbstractREPL, ::Any; backend_on_current_task::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:292
   [14] run_repl(::REPL.AbstractREPL, ::Any) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:288
   [15] (::Base.var"#806#808"{Bool,Bool,Bool,Bool})(::Module) at ./client.jl:399
   [16] #invokelatest#1 at ./essentials.jl:710 [inlined]
   [17] invokelatest at ./essentials.jl:709 [inlined]
   [18] run_main_repl(::Bool, ::Bool, ::Bool, ::Bool, ::Bool) at ./client.jl:383
   [19] exec_options(::Base.JLOptions) at ./client.jl:313
   [20] _start() at ./client.jl:506
```

I think the reason is that the random tangent for `Symbol`s is `DoesNotExist()` (rather than `nothing`), and this line does not catch it:
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/5111aef67238502926766094829239f9e99da6b0/src/testers.jl#L185

See also https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/113, perhaps we should stop using `nothing` altogether in the future.

mzgubic:
Closes #128 , closes #113

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -182,7 +182,7 @@ function test_rrule(
         @test ∂self === NO_FIELDS  # No internal fields
 
         # Correctness testing via finite differencing.
-        x̄s_is_dne = accumulated_x̄ .== nothing
+        x̄s_is_dne = isa.(accumulated_x̄, Union{Nothing, DoesNotExist})
         x̄s_fd = _make_j′vp_call(fdm, (xs...) -> f(xs...; fkwargs...), ȳ, xs, x̄s_is_dne)
         for (accumulated_x̄, x̄_ad, x̄_fd) in zip(accumulated_x̄, x̄s_ad, x̄s_fd)
             if accumulated_x̄ === nothing  # then we marked this argument as not differentiable
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -116,6 +116,14 @@ function test_frule(
 
         # TODO: remove Nothing when https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/113
         ẋs_is_ignored = isa.(ẋs, Union{Nothing, DoesNotExist})
+        if any(ẋs .== nothing)
+            Base.depwarn(
+                "test_frule(f, k ⊢ nothing) is deprecated, use " *
+                "test_frule(f, k ⊢ DoesNotExist()) instead for non-differentiable ks",
+                :test_frule
+            )
+        end
+
         # Correctness testing via finite differencing.
         dΩ_fd = _make_jvp_call(fdm, (xs...) -> f(deepcopy(xs)...; deepcopy(fkwargs)...), Ω, xs, ẋs, ẋs_is_ignored)
         check_equal(dΩ_ad, dΩ_fd; isapprox_kwargs...)
@@ -185,6 +193,14 @@ function test_rrule(
         # Correctness testing via finite differencing.
         # TODO: remove Nothing when https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/113
         x̄s_is_dne = isa.(accumulated_x̄, Union{Nothing, DoesNotExist})
+        if any(accumulated_x̄ .== nothing)
+            Base.depwarn(
+                "test_rrule(f, k ⊢ nothing) is deprecated, use " *
+                "test_rrule(f, k ⊢ DoesNotExist()) instead for non-differentiable ks",
+                :test_rrule
+            )
+        end
+
         x̄s_fd = _make_j′vp_call(fdm, (xs...) -> f(xs...; fkwargs...), ȳ, xs, x̄s_is_dne)
         for (accumulated_x̄, x̄_ad, x̄_fd) in zip(accumulated_x̄, x̄s_ad, x̄s_fd)
             if accumulated_x̄ === nothing  # then we marked this argument as not differentiable
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -182,6 +182,7 @@ function test_rrule(
         @test ∂self === NO_FIELDS  # No internal fields
 
         # Correctness testing via finite differencing.
+        # TODO: remove Nothing when https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/113
         x̄s_is_dne = isa.(accumulated_x̄, Union{Nothing, DoesNotExist})
         x̄s_fd = _make_j′vp_call(fdm, (xs...) -> f(xs...; fkwargs...), ȳ, xs, x̄s_is_dne)
         for (accumulated_x̄, x̄_ad, x̄_fd) in zip(accumulated_x̄, x̄s_ad, x̄s_fd)
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -75,7 +75,7 @@ end
 - `inputs` either the primal inputs `x`, or primals and their tangents: `x ⊢ ẋ`
    - `x`: input at which to evaluate `f` (should generally be set to an arbitary point in the domain).
    - `ẋ`: differential w.r.t. `x`, will be generated automatically if not provided
-     Non-differentiable arguments, such as indices, should have `ẋ` set as `nothing`.
+   Non-differentiable arguments, such as indices, should have `ẋ` set as `DoesNotExist()`.
 
 # Keyword Arguments
    - `output_tangent` tangent to test accumulation of derivatives against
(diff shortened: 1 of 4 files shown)


src/testers.jl
```diff
@@ -182,7 +182,7 @@ function test_rrule(
         @test ∂self === NO_FIELDS  # No internal fields
 
         # Correctness testing via finite differencing.
-        x̄s_is_dne = accumulated_x̄ .== nothing
+        x̄s_is_dne = isa.(accumulated_x̄, Union{Nothing, DoesNotExist})
```
oxinabox, line 185: this should have a todo.
And maybe a `Base.depwarn`?

mzgubic: `depwarn` for what? we kind of use both unfortunately

oxinabox:
I wonder if we should make the default `rand_tangent` for integer not be defined.
Since we can't tell if the user means a semantic integer  (like indexing) or a an integer as computation all convience (so it is beign treated as embeddeded inside Floats)
cf: https://github.com/JuliaDiff/ChainRulesCore.jl/issues/73

oxinabox:
Is this all it takes? Cool!
Can we leave this PR open until the follow up to ChainRules itself that deletes a bunch of unneeded `nothing`s has been tested?

mzgubic:
Yeah that's fine, this is not blocking for anything

src/testers.jl
oxinabox: Oh, I thought this was going to close 
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/113
?
so we would in future just use DoesNotExist ?

oxinabox: So we should `Base.Depwarn` if `nothing` is passed

mzgubic: Yeah actually I think you are right, but I didn't think about this from the start so need to check.

mzgubic:
Polished this a bit and tested on https://github.com/JuliaDiff/ChainRules.jl/pull/396, seems to work. If you could take another look that would be great.

> I wonder if we should make the default rand_tangent for integer not be defined.

Do you mean to **not** return `DoesNotExist()`, but rather complain and error?

src/testers.jl
oxinabox: If i am right then we should have a deprecation warning?

mzgubic: I've added one below, or am I missing something obvious?

mzgubic: @oxinabox not sure if you are happy with the current depwarn?

oxinabox:
> Do you mean to not return DoesNotExist(), but rather complain and error?

Yeah.

mzgubic:
Hmm, I don't know, would tend to think that erroring is better. I made a [separate issue], but the decision is independent of this PR I think. It does matter for the [ChainRules PR](https://github.com/JuliaDiff/ChainRules.jl/pull/396/files) however that I have used to test this PR.

mzgubic:
@oxinabox is there anything that needs to be done before this PR is approved and ready to go?

src/testers.jl
oxinabox: Cool

oxinabox:
> Can we leave this PR open until the follow up to ChainRules itself that deletes a bunch of unneeded nothings has been tested?

Just waiting til we are happy with https://github.com/JuliaDiff/ChainRules.jl/pull/396

Which I guess we are happy enough with t now?
Feels like we want to mess with a bunch of thigns and what type their `rand_tangent` is.
(Esp https://github.com/JuliaDiff/ChainRules.jl/pull/396#discussion_r609622110)
but I guess that can be a follow up PR

oxinabox (approved):

PR #131, opened 2021-03-29
fix tests for 1.6

mzgubic:
Fixes the tests failing on 1.6.

If i understand things correctly the `Test.Fail` object has changed its representation of its `orig_expr` field from `Expr` to `String`.

This fix is pretty terrible, any other ideas?


oxinabox:
see also #72 and discussion there

PR #140, opened 2021-04-20
Enable testing of rules with `@not_implemented`

devmotion:
Currently tests fail since `check_add!!` errors for `Composite`s with a field of type `NotImplemented`: the implementation of `+` for `Composite` in ChainRulesCore uses `elementwise_add` which is implemented with `+` but `+` is not defined for `NotImplemented`.

My suggestion would be to allow operations with `Composite`s with fields of type `NotImplemented` similar to how `muladd` is enabled for `NotImplemented`, i.e., `NotImplemented` wins always, _but_ to not allow general calculations with `+`. This could be achieved by defining
```julia
_add(x, y) = x + y
_add(x::NotImplemented, y) = x
_add(::NotImplemented, y::NotImplemented) = y
_add(x::NotImplemented, ::NotImplemented) = x
```
and using `_add` instead of `+` in `elementwise_add`. I already confirmed that this would fix the test errors.

### DIFF
--- a/src/check_result.jl
+++ b/src/check_result.jl
@@ -33,10 +33,10 @@ check_equal(x::Zero, y::Zero; kwargs...) = @test true
 check_equal(x::DoesNotExist, y::Nothing; kwargs...) = @test true
 check_equal(x::Nothing, y::DoesNotExist; kwargs...) = @test true
 
-check_equal(::ChainRulesCore.NotImplemented, x; kwargs...) = @test true
-check_equal(x, ::ChainRulesCore.NotImplemented; kwargs...) = @test true
+check_equal(::ChainRulesCore.NotImplemented, x; kwargs...) = @test_broken false
+check_equal(x, ::ChainRulesCore.NotImplemented; kwargs...) = @test_broken false
 function check_equal(::ChainRulesCore.NotImplemented, ::ChainRulesCore.NotImplemented; kwargs...)
-    return @test true
+    return @test_broken false
 end
 
 """
@@ -145,10 +145,10 @@ function _check_add!!_behaviour(acc, val; kwargs...)
 end
 
 # `+` is not defined for `NotImplemented`
-_check_add!!_behaviour(acc, ::ChainRulesCore.NotImplemented; kwargs...) = @test true
-_check_add!!_behaviour(::ChainRulesCore.NotImplemented, val; kwargs...) = @test true
+_check_add!!_behaviour(acc, ::ChainRulesCore.NotImplemented; kwargs...) = @test_broken false
+_check_add!!_behaviour(::ChainRulesCore.NotImplemented, val; kwargs...) = @test_broken false
 function _check_add!!_behaviour(
     ::ChainRulesCore.NotImplemented, ::ChainRulesCore.NotImplemented; kwargs...
 )
-    return @test true
+    return @test_broken false
 end
--- a/src/check_result.jl
+++ b/src/check_result.jl
@@ -35,10 +35,14 @@ check_equal(x::Nothing, y::DoesNotExist; kwargs...) = @test true
 
 # Checking equality with `NotImplemented` reports `@test_broken` since the derivative has intentionally
 # not yet been implemented
-check_equal(::ChainRulesCore.NotImplemented, x; kwargs...) = @test_broken false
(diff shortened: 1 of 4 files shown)


devmotion:
Hmm just thought, maybe one should use `@test_broken false` instead of `@test true` for `NotImplemented` to indicate more clearly that something should be fixed?

mzgubic:
Nice, thanks for adding this here as well. I would support the use of `@test_broken false` as you suggest (and maybe add a comment to the testset in this package that a broken test is expected.

With regards to the tests: Should we just add `+` method to `NotImplemented` as discussed in the first place? It does seem relatively harmless and would avoid the suggested workaround. Sorry for anticipating this when I was reviewing the previous PR!

src/check_result.jl
```diff
@@ -33,6 +33,12 @@ check_equal(x::Zero, y::Zero; kwargs...) = @test true
 check_equal(x::DoesNotExist, y::Nothing; kwargs...) = @test true
 check_equal(x::Nothing, y::DoesNotExist; kwargs...) = @test true
 
+check_equal(::ChainRulesCore.NotImplemented, x; kwargs...) = @test_broken false
```
oxinabox, line 36:
```suggestion
# Checking equality with `NotImplemented` reports `@test_broken` since the derivative has intentionalyl
# not yet been implemented
check_equal(::ChainRulesCore.NotImplemented, x; kwargs...) = @test_broken false
```

mzgubic:
Thanks for adding the update. I think it is good to go, but I do have a suggestion for an improvement. If you are busy with other things I can open an issue we can tackle it separately.

We use meta testing tools to test for errors:
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/0107180f94fcc4e45e7fd78bb0e04eece077c2dc/test/meta_testing_tools.jl#L97

Perhaps we could add a similar thing that would catch `broken` tests. This way, the current tests for `check_equal` would show up as passes.

mzgubic (approved):

src/check_result.jl
```diff
@@ -33,6 +33,12 @@ check_equal(x::Zero, y::Zero; kwargs...) = @test true
...
+check_equal(x, ::ChainRulesCore.NotImplemented; kwargs...) = @test_broken false
```
oxinabox, line 39: I wonder if throughout this section we should 
```suggestion
check_equal(x, y::ChainRulesCore.NotImplemented; kwargs...) = @test_broken x == y
```
Since seems like it gives a good message in testing outputs:

```
julia> @test_broken @not_implemented("derivative of foo(x,y) w.r.t y") == 1
Test Broken
  Expression: #= REPL[23]:1 =# @not_implemented("derivative of foo(x,y) w.r.t y") == 1
  ```

File: src/check_result.jl [hunk 3]
```diff
@@ -137,3 +143,12 @@ function _check_add!!_behaviour(acc, val; kwargs...)
     acc_mutated = deepcopy(acc)  # prevent this test changing others
     check_equal(add!!(acc_mutated, val), acc + val; kwargs...)
 end
+
+# `+` is not defined for `NotImplemented`
```
oxinabox: it is now, right?

oxinabox (approved):
Nice.
I have made some comments, but merge when you are happy.

src/check_result.jl
devmotion: Yes, I had noticed this as well, it is fixed now.

src/check_result.jl
devmotion: I am a bit worried that `x == y` might evaluate to true sometimes in the case below where `x` and `y` are both instances of `NotImplemented`. But maybe your suggestion is to only use `x == y` in the other cases where the types are different?

oxinabox: I didn't think about the case where they were both `NotImplemented`.
I wonder if that case should do `@test x == y`
I guess that case comes up when checking `add!!` ?
Not sure.

```julia
julia> @not_implemented("a") == @not_implemented("b")
false

julia> @not_implemented("a") == @not_implemented("a")
true
```

For other cases seems like better output is more good

devmotion: It's not completely clear to me if and how these non-constant tests could break but I didn't discover any problems with the more descriptive output in the tests, so I changed it.

PR #141, opened 2021-04-21
remove deprecated Symbol ⊢ nothing from tests

mzgubic:
Following up on https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/129

### DIFF
--- a/test/testers.jl
+++ b/test/testers.jl
@@ -298,11 +298,11 @@ end
         end
 
         @testset "test_frule" begin
-            test_frule(fsymtest, 2.5, :x ⊢ nothing)
+            test_frule(fsymtest, 2.5, :x)
         end
 
         @testset "test_rrule" begin
-            test_rrule(fsymtest, 2.5, :x ⊢ nothing)
+            test_rrule(fsymtest, 2.5, :x)
         end
     end
 


oxinabox:
Should we move this to `test/deprecated.jl` ?

mzgubic:
Hmm, I don't think so. The tests in `test/deprecated.jl` test deprecated functionality (`rrule_test`, `frule_test`). These tests some non-deprecated functionality using a deprecated syntax.

oxinabox:
Right but don't we want to test that that deprecated syntax still works until we decide to remove it?

mzgubic:
Possibly copy it then? So that it doesn't get deleted by mistake when deprecations are removed?

oxinabox:
When deprecations are removed we should be removing support for that also. So we should remove it.

Am I missing something? Should we have a call?

mzgubic:
Once deprecated `test_frule(fsymtest, 2.5, :x ⊢ nothing)` syntax is removed, we should still test `test_frule(fsymtest, 2.5, :x)` works, right?

oxinabox:
> Once deprecated `test_frule(fsymtest, 2.5, :x ⊢ nothing)` syntax is removed, we should still test `test_frule(fsymtest, 2.5, :x)` works, right?

Right.
and do we have a `test_frule(fsymtest, 2.5, :x ⊢ nothing)`   in `test/deprecated.jl` ?
I think we need to add one to 
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/653ae9880ed8cb97a3a7b9be5a7b2a9fa5e18651/test/deprecated.jl#L144

Or was `:x ⊢ nothing`  never really supported. Only ` :x ⊢ DoesNotExist()` and `(:x, nothing)` ?

oxinabox (approved):

PR #142, opened 2021-04-29
Automatically ignore `Composite`s with all `DoesNotExist()` for finite differencing

mzgubic:
Do we want this?

Closes https://github.com/JuliaDiff/FiniteDifferences.jl/issues/149

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.6.8"
+version = "0.6.9"
 
 [deps]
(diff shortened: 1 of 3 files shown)


oxinabox:
I don't think we want this.
Though we might want special handling for things that are `iszero` and that should be `iszero`.

mzgubic:
Yeah fair. On the other hand I think @willtebbutt  was reluctant to do a lot of special casing in `rand_tangent`

oxinabox:
I think special casing in rand_tangent is better.
Is a set of simpler methods.

If we did automatic simplification in it for structs it generated that were all DoesNotExist fields to become DoesNotExist then that would cover same as this but be nicer

PR #144, opened 2021-05-04
Change `check_thunking_is_appropriate` to a warning

mcabbott:
While thunking the sole nonzero argument may be unnecessary, it hardly seems the kind of grave sin for which waiting another hour for CI to tell you what else is actually wrong is the appropriate punishment. So this makes it a friendly warning instead.

For precedent, note that `Base.literal_pow(^, x, Val(1))` could safely be made an error since clearly you are wasting pixels if you write `x^1` and should ideally go and clean up. But it's nice that it isn't.

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -219,17 +219,15 @@ function test_rrule(
             end
         end
 
-        check_thunking_is_appropriate(x̄s_ad)
+        check_thunking_is_appropriate(f, x̄s_ad)
     end  # top-level testset
 end
 
-function check_thunking_is_appropriate(x̄s)
-    @testset "Don't thunk only non_zero argument" begin
-        num_zeros = count(x->x isa AbstractZero, x̄s)
-        num_thunks = count(x->x isa Thunk, x̄s)
-        if num_zeros + num_thunks == length(x̄s)
-            @test num_thunks !== 1
-        end
+function check_thunking_is_appropriate(f, x̄s)
+    num_zeros = count(x->x isa AbstractZero, x̄s)
+    num_thunks = count(x->x isa Thunk, x̄s)
+    if num_thunks == 1 && num_zeros + num_thunks == length(x̄s)
+        @warn "function has a thunk for its only non-zero argument, this works but is considered poor style" f
     end
 end
 


devmotion:
An argument against this change would be that it is very easy to miss the warning. AFAIK it would not show up in CI tests if you don't check the logs explicitly, would it?

mcabbott:
Indeed, you could miss one, if you don't read what's printed to the terminal. The worst-case scenario seems to be that you leave a supposedly zero-cost operation in your code. Which right now, if I understand right, describes every single thunk, since Zygote ignores them.

oxinabox:
>  it hardly seems the kind of grave sin for which waiting another hour for CI to tell you what else is actually wrong is the appropriate punishment.

It is inside a testset.
It shouldn't kill the whole CI run if it fails.

mcabbott:
Oh yes, maybe CI keeps running, and reports dozens of errors, maybe hundreds, from testsets so deeply nested that in my case they were all too wide to fit on the screen... and took a long time to track down. There's a lot of friction trying to add things, and so I think it's worth thinking hard about whether all of it is absolutely and completely necessary.

devmotion:
> The worst-case scenario seems to be that you leave a supposedly zero-cost operation in your code.

I was always convinced that thunks are not zero-cost operations and therefore, e.g., typically scalar operations should not be thunked (https://github.com/JuliaDiff/ChainRules.jl/pull/249).

mcabbott:
> e.g., typically scalar operations should not be thunked

OK, but will this test catch such things?

Perhaps testing for those would be too strict, as there could occasionally be expensive scalar operations which it would make sense to wrap.

A less joking example than `x^1` is that it's not uncommon to comment out parts of what you're working on, including arguments of functions. It seems pretty surprising to me that a valid 2-argument function should cause an error when I comment out the 2nd argument temporarily, in order to test first whether I got the other argument right. That's a weird nonlocal effect.

oxinabox:
> Oh yes, maybe CI keeps running, and reports dozens of errors, maybe hundreds, from testsets so deeply nested that in my case they were all too wide to fit on the screen... and took a long time to track down. There's a lot of friction trying to add things, and so I think it's worth thinking hard about whether all of it is absolutely and completely necessary.

We should work on making the output more useful and clear.
But I don't think removing tests that catch (what is currently deemed to be minor) actual user errors is the way to get it.
Maybe we could have modes that are set on a global level? so can disable tests on things you don't care about while working on the PR. 
(though in effect you can, as you did, achieve this with `@eval ChainRulesTestUtils check_thunking_is_appropriate(false)`)

Warnings get lost easily.
Especially since we test a bunch of deprecated things to make sure they still work.

mcabbott:
Maybe. Global test modes you have to read the manual to understand also sounds pretty complicated. 

What I'm trying to say is that there's a lot, a *lot*, of friction in trying to add things. And as a result lots of things don't get added, people who could fix bugs give up on making the PR ever work. I think it would pay to make a much bigger attack on the degree of nested complication in these tests. This is just a tiny one which cost me a few hours.

And the worst case here, if a warning gets missed, and no reviewer has memorised the manual deeply enough to spot it, is that next week someone sees it and makes a quick fix. It doesn't affect correctness. Nothing is benchmarked to within 1% or whatever.

devmotion:
You can also use `@not_implemented` to skip tests for one differential instead of commenting out the argument 🙂

mcabbott:
> You can also use @not_implemented to skip tests for one differential

Like this? 
```
julia> ChainRulesTestUtils.check_thunking_is_appropriate((Zero(),@thunk(1)))
Don't thunk only non_zero argument: Test Failed at /Users/me/.julia/packages/ChainRulesTestUtils/bDd51/src/testers.jl:231
  Expression: num_thunks !== 1
   Evaluated: 1 !== 1
Stacktrace:
```
Note, aside, that `1 !== 1` was a pretty confusing thing, why `===`? I spent a while wondering whether an `Int32` had got in there, since those print the same. 
```
 [1] record(ts::Test.DefaultTestSet, t::Union{Test.Error, Test.Fail})
   @ Test ~/.julia/dev/julia/usr/share/julia/stdlib/v1.7/Test/src/Test.jl:910
 [2] do_test(result::Test.ExecutionResult, orig_expr::Any)
   @ Test ~/.julia/dev/julia/usr/share/julia/stdlib/v1.7/Test/src/Test.jl:620
Test Summary:                      | Fail  Total
Don't thunk only non_zero argument |    1      1
ERROR: Some tests did not pass: 0 passed, 1 failed, 0 errored, 0 broken.

julia> ChainRulesTestUtils.check_thunking_is_appropriate((Zero(),@not_implemented(@thunk(1))))
ERROR: MethodError: Cannot `convert` an object of type Thunk{var"#9#10"} to an object of type String
Closest candidates are:
  convert(::Type{String}, ::String) at essentials.jl:233
  convert(::Type{T}, ::T) where T<:AbstractString at strings/basic.jl:231
  convert(::Type{T}, ::AbstractString) where T<:AbstractString at strings/basic.jl:232

julia> ChainRulesTestUtils.check_thunking_is_appropriate((Zero(),@not_implemented(1)))
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type String
Closest candidates are:
  convert(::Type{String}, ::String) at essentials.jl:233
  convert(::Type{T}, ::T) where T<:AbstractString at strings/basic.jl:231
  convert(::Type{T}, ::AbstractString) where T<:AbstractString at strings/basic.jl:232
  ...
```
And, again, this is one more layer of complexity. Why do I need a syntax transformation to comment out an argument?

oxinabox:
> Note, aside, that 1 !== 1 was a pretty confusing thing, why ===? I spent a while wondering whether an Int32 had got in there, since those print the same.

yeah, that is probably not needed, should remove that.
They are always both `Int` but it's not like tht matters.
(technically there is a performance gain but it is super tiny and we do waay slower things)

devmotion:
Maybe I am biased here but to me it seems the natural approach if you don't care about one of the differentials (at least initially) is to use `@not_implemented("skipped for now")` instead of a proper differential. That's what the macro is designed for (https://juliadiff.org/ChainRulesCore.jl/stable/writing_good_rules.html#Use-@not_implemented-appropriately). In some way this seems almost more "correct" than commenting out arguments of the function of interest.

mcabbott:
> In some way this seems almost more "correct" 

Yes I can see that. Maybe it's a nice feature for experts. (Although I remain surprised it doesn't print a warning or anything, like `@test_broken` does.)

But I didn't even know this existed, and I've contributed quite a bit. 

Breaking the ability to do simple things in the language, like convert a valid 3-argument function to a valid 2-argument function with the `#` key, is pretty confusing! And if you're testing informally by calling Zygote on your new code & comparing to ForwardDiff, you won't notice, because it works fine.

nickrobinson251:
> What I'm trying to say is that there's a lot, a lot, of friction in trying to add things

It sounds like we should dedicate some time and effort to improving this! 

fwiw, on first look (having read the discussion), I don't think disabling these particular tests are a good solution... But i would definite welcome more suggestions, and we should definitely take a look at what changes could be made and be willing to make some compromises in favour of making it easier to contribute (possibly after that this one will look like a good option, but right now removing a test doesn't seem a good solution to a confusing failure and verbose output.

mcabbott:
To reduce complexity, I think it's probably necessary to be willing to cut some unimportant things. I couldn't think of a smaller change to start with.

Bigger changes, it's pretty hard to track down which `rrule_test` failed, so that you can run it in isolation locally. Because the actual `@test`s are deep inside that function, so they don't have useful line numbers. And because they're all in deeply nested loops of parameters, you get 100 failures but all you want is one example to debug. I appreciate the desire to test things well but it's maddeningly hard to figure out what's going wrong. Base is easier!

I proposed elsewhere something like `@rrule_test  fun(Rand(4,4), dims=1)  atol=0.1` -- using a macro so that it knows what line number you called it from, and so that you don't have to remember how to work the multiple levels of keyword arguments. And some shorthand like `Rand(4,4)` meaning "please try this on a Matrix, and an Adjoint matrix of integers, and a Diagonal, and some complex numbers" and tell me which one failed. Ideally by printing exactly `fun(UpperTriangular(rand(4,4) .+ im, dims=1)` or something. But this might be a big project.

mcabbott:
What I forgot, from the PR:

* My original case was one `InplaceableThunk`, which was causing type instabilities, so I commented out the in-place part, leaving the `@thunk`. So you have to know that it's only one kind of thunk not the other which isn't allowed to be alone. 
* The proposal of The https://github.com/JuliaDiff/ChainRulesCore.jl/issues/274 is to make `@thunk` generate the Inplace part automatically, at least for simple enough thunks. Then you will need also to remember this, and know whether it's simple enough. 

Anyway, I close this at 3 votes to 1.

nickrobinson251:
> My original case was one InplaceableThunk, which was causing type instabilities, so I commented out the in-place part, leaving the @thunk. So you have to know that it's only one kind of thunk not the other which isn't allowed to be alone.

Oh i think that's a bug. Here's a PR to fix that: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/145

Or if that's intended we should document that much better (we can discuss in https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/145)

PR #145, opened 2021-05-05
RFC: Check only non-zero argument not an InpaceableThunk?

nickrobinson251:
Raised by @mcabbott at https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/144#issuecomment-832361804

> My original case was one InplaceableThunk, which was causing type instabilities, so I commented out the in-place part, leaving the @thunk. So you have to know that it's only one kind of thunk not the other which isn't allowed to be alone.

Is this intended? If no, this PR fixes it. If it is intended for `Thunk` to be treated differently than `InpaceableThunk` we should probably have docs about that...

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -134,7 +134,6 @@ function test_frule(
 end
 
 
-
 """
     test_rrule(f, inputs...; kwargs...)
 
@@ -225,8 +224,8 @@ end
 
 function check_thunking_is_appropriate(x̄s)
     @testset "Don't thunk only non_zero argument" begin
-        num_zeros = count(x->x isa AbstractZero, x̄s)
-        num_thunks = count(x->x isa Thunk, x̄s)
+        num_zeros = count(x -> x isa AbstractZero, x̄s)
+        num_thunks = count(x -> x isa AbstractThunk, x̄s)
         if num_zeros + num_thunks == length(x̄s)
             @test num_thunks !== 1
         end


nickrobinson251:
Looking back at history, there was a comment (https://github.com/JuliaDiff/ChainRulesTestUtils.jl/commit/ef5f72841ceaf38b99d2528ab0cacc09e2a38fa7#diff-988b853ad3c22aebac9421d3a31e60f9330789df43bccfb9fb75365048398fbaL260)

> single adjoint should not be `Thunk`ed. InplaceableThunk is fine.

So i guess the code is as intended... but phwoar that's a gotcha isn't it? How can we document that better?

oxinabox:
>      So i guess the code is as intended... but phwoar that's a gotcha isn't it? How can we document that better

Why is it a gotcha?

There is no benifit to the only return being a `Thunk` since it will be unthunked if it is used (which it almost certainly will be)
But an `InplaceableThunk` does have a benifit: it can be applied inplace.

nickrobinson251:
yeah... maybe this isn't an issue

I just have a lot of sympathy for e.g. `@mcabbot` running into so much friction when trying to contribute, and hitting new issue when trying to debug some other issue

maybe like #144, this isn't a good fix...

nickrobinson251:
yeah... i think like #144  this isn't the right fix... but we do need to do something to make contributing easier.

here's something in that direction for discussion: #146

PR #155, opened 2021-05-28
replace `NO_FIELDS` with `NoTangent()`

mzgubic:
Follows up https://github.com/JuliaDiff/ChainRulesCore.jl/pull/358

Breaking because it requires the `dself` to be `NoTangent` rather than `NO_FIELDS`, i.e. `ZeroTangent`.


oxinabox (approved):

mzgubic:
Do you have an idea of why the Documenter job would not find ChainRulesCore0.10? All the other jobs do

PR #158, opened 2021-05-28
Avoid nesting testsets in test_rule

### LINKED ISSUE #146: How to improve test report/logs for when tests fails?
Here's an example of what the test report looks like when some `test_rrule` tests fail (from https://github.com/JuliaDiff/ChainRulesTestUtils.jl/runs/2508421035?check_suite_focus=true)
 
 
![Screenshot 2021-05-05 at 12 07 45](https://user-images.githubusercontent.com/13448787/117132305-97e70280-ad9a-11eb-9d3a-d2969a70088f.png)
![Screenshot 2021-05-05 at 12 08 06](https://user-images.githubusercontent.com/13448787/117132324-9e757a00-ad9a-11eb-8cbc-c7e9358b0d23.png)

I find this test report really difficult to read.

A good thing about this is that it contains a lot of potentially relevant information, e.g. 
```
          test_rrule: sort at ([0.23603334566204692, 0.34651701419196046, 0.3127069683360675, 0.00790928339056074, 0.4886128300795012, 0.21096820215853596, 0.951916339835734, 0.9999046588986136, 0.25166218303197185, 0.9866663668987996],)                                                                                                                 |     4     1      5
            Don't thunk only non_zero argument                                                                                                                                                                                                                                                                                                                |           1      1
```
Bad things about printing all this info( like `sort at ([0.23603334566204692, 0.34651701419196046, ...],)`), in my opinion:
- this info  is in the wrong place to be helpful
  - I want to see this info well digging into _why_ something failed (when scrolling up back through the logs to investigate the actual failure e.g. 
  ```
  Don't thunk only non_zero argument: Test Failed at /home/runner/work/ChainRulesTestUtils.jl/ChainRulesTestUtils.jl/src/testers.jl:230
  Expression: num_thunks !== 1
   Evaluated: 1 !== 1`
    ```
    (and as it turns out, this info isn't relevant to the failures `num_thunks !== 1`)
- this obscures the info i about _what_ test failed (e.g. the `rrule` for `sort`, etc.), _how many_ tests failed, and any _patterns_ in the failures 
  - e.g. it's hard to see that 23 `getindex` tests fail
  - e.g. it's hard to see it's all failures for `rrule`s and all fail the same `Don't thunk only non_zero argument` test
- this info feels overwhelming (because it breaks the pretty printing / due to the amount of info and the pretty test report printing look how far to the right you need to scroll to see the numbers in the first snippet above)

Here's the raw logs with the window streched so each test failure is it's own line... this makes the nesting of tests clear at least, but it'd be a 3 day trek east to find the assocaited number of failures
![Screenshot 2021-05-05 at 12 22 50](https://user-images.githubusercontent.com/13448787/117134084-05942e00-ad9d-11eb-87ce-62978aec7e6a.png)

---

The question is: how can we improve this? i.e. make it as easy as possible both to interpretting test failures and knowing what exactly went wrong.

some sugegstions:
- move some of this info out of testset names and into log messages 
  - unfortunately Test stdlib doesn’t have messages you can just attach to failures...  and printing a log message for every test may be even more overwhelming...
- reduce the use of long random numbers (`sort at ([0.23603334566204692, 0.34651701419196046, ...],)`) which pushes info to the right
  - e.g. replace them with numbers that have have a short decimal representation (e.g. `rand(-2:0.1:2)`) (suggested by @oxinabox )
- others options??

oxinabox:
Combined with the other PRs i have open this should close #146.

We don't nest testsets any more, except in `test_scalar` (and I don't think thats much of a problem just cos scalars rarely fail their pullbacks are simple).
Instead we use a hack to make `@test` print a message on failure.
and then we plumb that around as needed.

Example of one of the place we would nest a test set before. Mocking up the outer set just for this demo
```julia
julia> @testset "rrule for foo at bar" begin
           check_equal(Tangent{Foo}(1, 2), Tangent{Foo}(1, 3))
       end
rrule for foo at bar: Test Failed at /Users/oxinabox/JuliaEnvs/ChainRulesWorld/ChainRules.jl/dev/ChainRulesTestUtils/src/check_result.jl:25
  Problem:  Foo.2
   Evaluated: isapprox(2, 3)
Stacktrace:
 [1] check_equal(actual::Int64, expected::Int64, msg::String; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ChainRulesTestUtils ~/JuliaEnvs/ChainRulesWorld/ChainRules.jl/dev/ChainRulesTestUtils/src/check_result.jl:25
 [2] check_equal
   @ ~/JuliaEnvs/ChainRulesWorld/ChainRules.jl/dev/ChainRulesTestUtils/src/check_result.jl:25 [inlined]
 [3] check_equal(actual::Tangent{Foo, Tuple{Int64, Int64}}, expected::Tangent{Foo, Tuple{Int64, Int64}}, msg::String; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ChainRulesTestUtils ~/JuliaEnvs/ChainRulesWorld/ChainRules.jl/dev/ChainRulesTestUtils/src/check_result.jl:92
 [4] check_equal (repeats 2 times)
   @ ~/JuliaEnvs/ChainRulesWorld/ChainRules.jl/dev/ChainRulesTestUtils/src/check_result.jl:86 [inlined]
 [5] macro expansion
   @ REPL[189]:2 [inlined]
 [6] macro expansion
   @ /usr/local/src/julia/julia-1.6/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
 [7] top-level scope
   @ REPL[189]:2
Test Summary:        | Pass  Fail  Total
rrule for foo at bar |    1     1      2
```

vs before:
```julia
julia> @testset "rrule for foo at bar" begin
           check_equal(Tangent{Foo}(1, 2), Tangent{Foo}(1, 3))
       end
Foo.2: Test Failed at /Users/oxinabox/.julia/packages/ChainRulesTestUtils/0XyYu/src/check_result.jl:19
   Evaluated: isapprox(2, 3)
Stacktrace:
 [1] check_equal(actual::Int64, expected::Int64; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ChainRulesTestUtils ~/.julia/packages/ChainRulesTestUtils/0XyYu/src/check_result.jl:19
 [2] check_equal
   @ ~/.julia/packages/ChainRulesTestUtils/0XyYu/src/check_result.jl:19 [inlined]
 [3] macro expansion
   @ ~/.julia/packages/ChainRulesTestUtils/0XyYu/src/check_result.jl:84 [inlined]
 [4] macro expansion
   @ /usr/local/src/julia/julia-1.6/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1226 [inlined]
 [5] check_equal(actual::Tangent{Foo, Tuple{Int64, Int64}}, expected::Tangent{Foo, Tuple{Int64, Int64}}; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ChainRulesTestUtils ~/.julia/packages/ChainRulesTestUtils/0XyYu/src/check_result.jl:83
Test Summary:        | Pass  Fail  Total
rrule for foo at bar |    1     1      2
  Foo.1              |    1            1
  Foo.2              |          1      1
ERROR: Some tests did not pass: 1 passed, 1 failed, 0 errored, 0 broken.
```

which doesn't look like much and i guess probably wasn't a great example (someone should find and try some others)

But it is nice that the error now at the top is about what is going on
now: `rrule for foo at bar: Test Failed at /Users/oxinabox/JuliaEnvs/ChainRulesWorld/ChainRules.jl/dev/ChainRulesTestUtils/src/check_result.jl:25`
before: `Foo.2: Test Failed at /Users/oxinabox/.julia/packages/ChainRulesTestUtils/0XyYu/src/check_result.jl:19`

In general having the `testset` breakdown sounds nice.
But it isn't actually useful
What you really want is the error message:
```
  Problem:  Foo.2
   Evaluated: isapprox(2, 3)
 ```
+ the line in the file that is within `test/...` which triggered it (so you can see exactly how it was configured).
Everything else is kinda noise.
but that info is no where near the testset summary at the end.

It also tweaks the content of the testset title that was set in #154 so that it also prints just primal type even if  you passed in PrimalAndTangent. (this avoids printingthat long `ChainrulesCore.PrimalAndTangent{Float64, Float64}` type).

### DIFF
--- a/src/output_control.jl
+++ b/src/output_control.jl
@@ -41,6 +41,9 @@ julia> @test_msg a sum(1:1000) < 10;
 ```
 """
 macro test_msg(msg, ex, kws...)
+    # This code is basically a evil hack that accesses the internals of the Test stdlib.
+    # Code below is based on the `@test` macro definition as it was in Julia 1.6.
+    # https://github.com/JuliaLang/julia/blob/v1.6.1/stdlib/Test/src/Test.jl#L371-L376
     Test.test_expr!("@test_msg msg", ex, kws...)
 
     result = Test.get_test_result(ex, __source__)
--- a/src/output_control.jl
+++ b/src/output_control.jl
@@ -1,0 +1,59 @@
+# Test.get_test_result generates code that uses the following so we must import them
+using Test: Returned, Threw, eval_test
+
+"A cunning hack to carry extra message along with the original expression in a test"
+struct ExprAndMsg
+    ex
+    msg
+end
+
+"""
+    @test_msg msg condion kws...
+
+This is per `Test.@test condion kws...` except that if it fails it also prints the `msg`.
+If `msg==""` then this is just like `@test`, nothing is printed
+
+### Examles
+```julia
+julia> @test_msg "It is required that the total is under 10" sum(1:1000) < 10;
+Test Failed at REPL[1]:1
+  Expression: sum(1:1000) < 10
+  Problem: It is required that the total is under 10
+   Evaluated: 500500 < 10
+ERROR: There was an error during testing
+
+
+julia> @test_msg "It is required that the total is under 10" error("not working at all");
+Error During Test at REPL[2]:1
+  Test threw exception
+  Expression: error("not working at all")
+  Problem: It is required that the total is under 10
+  "not working at all"
+  Stacktrace:
+
+julia> a = "";
+
+julia> @test_msg a sum(1:1000) < 10;
+  Test Failed at REPL[153]:1
+    Expression: sum(1:1000) < 10
+     Evaluated: 500500 < 10
+  ERROR: There was an error during testing
+```
+"""
+macro test_msg(msg, ex, kws...)
+    Test.test_expr!("@test_msg msg", ex, kws...)
+
+    result = Test.get_test_result(ex, __source__)
+    return :(Test.do_test($result, $ExprAndMsg($(string(ex)), $(esc(msg)))))
+end
+
+function Base.print(io::IO, x::ExprAndMsg)
+    print(io, x.ex)
+    !isempty(x.msg) && print(io, "\n  Problem: ", x.msg)
+end
+
+
+### helpers for printing in log messages etc
+_string_typeof(x) = string(typeof(x))
+_string_typeof(xs::Tuple) = join(_string_typeof.(xs), ",")
+_string_typeof(x::PrimalAndTangent) = _string_typeof(primal(x))  # only show primal
--- a/src/ChainRulesTestUtils.jl
+++ b/src/ChainRulesTestUtils.jl
@@ -17,9 +17,12 @@ export TestIterator
 export check_equal, test_scalar, test_frule, test_rrule, generate_well_conditioned_matrix
 export ⊢
 
+
 include("generate_tangent.jl")
 include("data_generation.jl")
 include("iterator.jl")
+
+include("output_control.jl")
 include("check_result.jl")
 
 include("finite_difference_calls.jl")
--- a/src/check_result.jl
+++ b/src/check_result.jl
@@ -4,50 +4,56 @@
 # Note that this must work well both on Differential types and Primal types
 
 """
-    check_equal(actual, expected; kwargs...)
+    check_equal(actual, expected, [msg]; kwargs...)
 
 `@test`'s  that `actual ≈ expected`, but breaks up data such that human readable results
 are shown on failures.
 Understands things like `unthunk`ing `ChainRuleCore.Thunk`s, etc.
+
+If provided `msg` is printed on a failure. Often additional items are appended to `msg` to
+give bread-crumbs into nested structures.
+
 All keyword arguments are passed to `isapprox`.
 """
 function check_equal(
     actual::Union{AbstractArray{<:Number},Number},
-    expected::Union{AbstractArray{<:Number},Number};
+    expected::Union{AbstractArray{<:Number},Number},
+    msg="",
+    ;
     kwargs...,
 )
-    @test isapprox(actual, expected; kwargs...)
+    @test_msg msg isapprox(actual, expected; kwargs...)
 end
 
 for (T1, T2) in ((AbstractThunk, Any), (AbstractThunk, AbstractThunk), (Any, AbstractThunk))
-    @eval function check_equal(actual::$T1, expected::$T2; kwargs...)
-        return check_equal(unthunk(actual), unthunk(expected); kwargs...)
+    @eval function check_equal(actual::$T1, expected::$T2, msg=""; kwargs...)
+        return check_equal(unthunk(actual), unthunk(expected), msg; kwargs...)
     end
 end
 
-check_equal(::ZeroTangent, x; kwargs...) = check_equal(zero(x), x; kwargs...)
-check_equal(x, ::ZeroTangent; kwargs...) = check_equal(x, zero(x); kwargs...)
-check_equal(x::ZeroTangent, y::ZeroTangent; kwargs...) = @test true
+check_equal(::ZeroTangent, x, msg=""; kwargs...) = check_equal(zero(x), x, msg; kwargs...)
+check_equal(x, ::ZeroTangent, msg=""; kwargs...) = check_equal(x, zero(x), msg; kwargs...)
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: src/output_control.jl [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+function Base.print(io::IO, x::ExprAndMsg)
+    print(io, x.ex)
+    !isempty(x.msg) && print(io, "\n  Problem: ", x.msg)
+end
+
+
+### helpers for printing in log messages etc
+_string_typeof(x) = string(typeof(x))
+_string_typeof(xs::Tuple) = join(_string_typeof.(xs), ",")
+_string_typeof(x::PrimalAndTangent) = _string_typeof(primal(x))  # only show primal
```
mzgubic, line 59:
```suggestion
_string_typeof(x::PrimalAndTangent) = "$(_string_typeof(primal(x))) ⊢  $(_string_typeof(tangent(x)))"
```
Are we ever interested in the tangent type? Probably not?

src/output_control.jl
```diff
@@ -0,0 +1,59 @@
...
+    Expression: sum(1:1000) < 10
+     Evaluated: 500500 < 10
+  ERROR: There was an error during testing
+```
+"""
+macro test_msg(msg, ex, kws...)
+    Test.test_expr!("@test_msg msg", ex, kws...)
+
+    result = Test.get_test_result(ex, __source__)
+    return :(Test.do_test($result, $ExprAndMsg($(string(ex)), $(esc(msg)))))
```
mzgubic, line 47: Is using undocumented functions a code smell of some sort? I think it indicates that either:
1) `Test` package should be easier to extent/modify
or
2) We are not achieving our goal in the right way with the tools available

I am leaning towards 1) from other experience with `Test`. On the other hand, I don't fully understand this PR (namely the `test_msg` macro). What do `get_test_result` and `do_test` actually do?

mzgubic (commented):
Needs a version bump.

src/output_control.jl
```diff
@@ -0,0 +1,59 @@
+# Test.get_test_result generates code that uses the following so we must import them
+using Test: Returned, Threw, eval_test
```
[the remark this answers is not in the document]
oxinabox: see the comment right above this line.

src/output_control.jl
oxinabox: yeah i considered that. Note that that would display things like `Vector{Float64}  ⊢  NoTangent()` which is a bit of a play on words since the  ⊢ operator doesn't actually accept types.

I think we can leave it as is (displaying just primals) for now, and consider adding that later.

src/output_control.jl
oxinabox: AFAIK it is 1) I will add a comment. I don't really understand what they do either, just that this is the thing that is needed.
It is kinda copied directly from the Test stdlib.
This whole thing with `ExprAndMsg` is a huge hack.

src/output_control.jl
```diff
@@ -0,0 +1,59 @@
...
+julia> a = "";
+
+julia> @test_msg a sum(1:1000) < 10;
+  Test Failed at REPL[153]:1
+    Expression: sum(1:1000) < 10
+     Evaluated: 500500 < 10
+  ERROR: There was an error during testing
+```
+"""
+macro test_msg(msg, ex, kws...)
```
oxinabox, line 43:
```suggestion
macro test_msg(msg, ex, kws...)
    # This code is basically a evil hack that accesses the internals of the Test stdlib.
    # Code below is based on the `@test` macro definition as it was in Julia 1.6.
    # https://github.com/JuliaLang/julia/blob/v1.6.1/stdlib/Test/src/Test.jl#L371-L376
```

src/output_control.jl
oxinabox: comment added

mzgubic: Yeah, fair enough. LGTM just needs a version bump

mzgubic (approved):

PR #162, opened 2021-06-02
Add deprecation warnings to isapprox

oxinabox:
Even though we don't have a direct replacement, still good to give a warning.


mzgubic (approved):
LGTM modulo version bump

src/deprecated.jl
```diff
@@ -2,38 +2,80 @@
 # We are silently deprecating them as there is no alternative we are providing
```
nickrobinson251, line 2:
```suggestion
```

nickrobinson251 (approved):

PR #163, opened 2021-06-02
implement `test_approx` for `Eigen`

simeonschaub:
This is needed for https://github.com/JuliaDiff/ChainRules.jl/pull/431.

### DIFF
--- a/src/check_result.jl
+++ b/src/check_result.jl
@@ -24,6 +24,13 @@ function test_approx(
     @test_msg msg isapprox(actual, expected; kwargs...)
 end
 
+# Comparing eigen factoriztions is tricky, because the order and normalization of eigen 
+# vectors is not unique. Therefore we just calculate the original matrix and compare that
+# instead.
+function test_approx(actual::Eigen, expected::Eigen, msg=""; kwargs...)
+    return test_approx(Matrix(actual), Matrix(expected), msg; kwargs...)
+end
+
 for (T1, T2) in ((AbstractThunk, Any), (AbstractThunk, AbstractThunk), (Any, AbstractThunk))
     @eval function test_approx(actual::$T1, expected::$T2, msg=""; kwargs...)
         return test_approx(unthunk(actual), unthunk(expected), msg; kwargs...)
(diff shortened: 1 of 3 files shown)


simeonschaub:
@sethaxen Is this strong enough for the eigen tests in ChainRules? Or do we actually want to require the eigenvectors to be approximately equal as well?

src/check_result.jl
```diff
@@ -24,6 +24,13 @@ function test_approx(
     @test_msg msg isapprox(actual, expected; kwargs...)
 end
 
+# Comparing eigen factoriztions is tricky, because the order and normalization of eigen 
+# vectors is not unique. Therefore we just calculate the original matrix and compare that
+# instead.
+function test_approx(actual::Eigen, expected::Eigen, msg=""; kwargs...)
```
oxinabox, line 30: I wonder if we should do this for all subtypes of `Factorization` ?

simeonschaub: Hm, yes, maybe? Although I am not really familiar with those, so perhaps we should wait until an actual use case comes up?

oxinabox: Fair enough.

oxinabox:
I will let @sethaxen  give final approval on this, as I don;t know how things work with `Eigen`

sethaxen:
This is to be used to test that the primal part of `frule` and `rrule` returns the same decomposition as the primal function, right? In that case, I think we do care whether the eigenvalues and eigenvectors match. But also, it should not be hard to make them match, right, since the `eigen` rules call `eigen` internally? Perhaps I'm missing something.

simeonschaub:
Hmm, yes you might be right. Looking at the failures in https://github.com/JuliaDiff/ChainRules.jl/runs/2769420821?check_suite_focus=true#step:6:282 for example again, it looks like the values shown do actually match precisely, any ideas what could cause this? There shouldn't be any NaN's occuring in those tests, right? Even if there would be degeneracy, that shouldn't cause the results to be non-deterministic no? I can look into this some more if you don't have any ideas either.

sethaxen:
> it looks like the values shown do actually match precisely, any ideas what could cause this? There shouldn't be any NaN's occuring in those tests, right?

I'm not certain. But I do see some `undef`'s in that failed test. Is it possible a matrix has undefined values?

simeonschaub:
I only see that for the arrays with 0 elements, but I could be wrong.

simeonschaub:
Ah, wait! It seems like `Eigen now also contains condition numbers which are just random noise. That's probably why tests fail.

PR #166, opened 2021-06-02
test arbitrary functions with `f/rrule`-like API

### LINKED ISSUE #114: *rule_test should take a frule_like/rrule_like function so we can test AD systems
In order to make `frule_test` and `rrule_test` more generally useful for testing AD systems,
we should make them take a keyword argument that specifies what function that what to use.
This function would be used in the place of the default `frule`/`rrule`.

It would need to match the API of `frule` and `rrule`.
Just like the function that needs to be provided for https://github.com/JuliaDiff/ChainRulesCore.jl/issues/68

Then we would be able to use these tools to test AD systems

### LINKED ISSUE #173: Testing Rules that use RuleConfigs
Need for testing https://github.com/JuliaDiff/ChainRules.jl/pull/441#issuecomment-857986594

We could use
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/166
passing in `rrule_f = (args...; kwargs...) -> rrule(ADviaRuleConfig(args...; kwargs...))`

But probably we would like to have `test_rrule`'s positional arguments continue matching `rrule`'s.
and so we should make another PR that makes it take a optional first argument that is the config.
(and maybe by default it will be this `ADviaRuleConfig` ?)
If we do that https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/166 would want to be updated to also accept a config argument. which end user providing the `rrule_f` can ignore.
Or for testing their AD if they are also testing there config they can pass their config and set `rrule_f=rrule_via_ad` ?

mzgubic:
Closes #114
Closes #173

To do:
- [x] tests for passing in `xrule_f`
- [x] tests for config
- [x] documentation
- [x] [implement an actual `rrule_f` example in Zygote](https://github.com/FluxML/Zygote.jl/pull/987) to test real examples
- [x] think about matching requirements for `f/rrule_via_ad`
- [x] decide whether to define `test_approx(NoTangent, x) = test_approx(zero(x), x)`
- [x] wait for [configs](https://github.com/JuliaDiff/ChainRulesCore.jl/pull/363) to be merged and tagged, and update compat

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
-ChainRulesCore = "0.10"
+ChainRulesCore = "0.10.4"
 Compat = "3"
 FiniteDifferences = "0.12.12"
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.7.0"
+version = "0.7.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.7.1"
+version = "0.7.3"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.7.8"
+version = "0.7.9"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -145,6 +145,24 @@ In particular, when specifying the input tangents to [`test_frule`](@ref) and th
 As these tangents are used to seed the derivative computation.
 Inserting inappropriate zeros can thus hide errors.
 
+## Testing higher order functions
+
+Higher order functions, such as `map`, take a function (or a functor) `f` as an argument.
+`f/rrule`s for these functions call back into AD to compute the `f/rrule` of `f`.
+To test these functions, we use a dummy AD system, which simply calls the appropriate rule for `f` directly.
+For this reason, when testing `map(f, collection)`, the rules for `f` need to be defined.
+The `RuleConfig` for this dummy AD system is the default one, and does not need to be provided.
+```julia
+test_rrule(map, x->2x [1, 2, 3.]) # fails, because there is no rrule for x->2x
+
+mydouble(x) = 2x
+function ChainRulesCore.rrule(::typeof(mydouble), x)
+    mydouble_pullback(ȳ) = (NoTangent(), ȳ)
+    return mydouble(x), mydouble_pullback
+end
+test_rrule(map, mydouble, [1, 2, 3.]) # works
+```
+
 ## Testing AD systems
 
 The gradients computed by AD systems can be also be tested using `test_rrule`.
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -122,6 +122,46 @@ In particular, when specifying the input tangents to [`test_frule`](@ref) and th
 As these tangents are used to seed the derivative computation.
 Inserting inappropriate zeros can thus hide errors.
 
+## Testing AD systems
+
+The gradients computed by AD systems can be tested using `test_rrule` by providing an `rrule_f`/`frule_f` keyword argument.
+`rrule_f` is a function that wraps the gradient computation by an AD system in the same API as the `rrule`.
+For example, let's say we have a complicated function
+
+```julia
+function complicated(x, y)
+    return do(x + y) + some(x) * hard(y) + maths(x * y)
+end
+```
+
+that we do not know an `rrule` for, and we want to check whether the gradients provided by the AD system are correct.
+
+Firstly, we need to define an `rrule`-like function which wraps the gradients computed by AD.
+
+Let's say the AD package uses some custom differential types and does not provide a gradient w.r.t. the function itself.
+In order to make the pullback compatible with the `rrule` API we need to add a `NoTangent()` to represent the differential w.r.t. the function itself.
+We also need to transform the `ChainRules` differential types to the custom types (`cr2custom`) before feeding the `Δ` to the AD-generated pullback, and back to `ChainRules` differential types when returning from the `rrule` (`custom2cr`).
+
+```julia
+function ad_rrule(f::Function, args...)
+    y, ad_pullback = ADSystem.pullback(f, args...)
+    function rrulelike_pullback(Δ)
+        diffs = custom2cr(ad_pullback(cr2custom(Δ)))
+        return NoTangent(), diffs...
+    end
+        
+    return y, rrulelike_pullback
+end
+
+custom2cr(differential) = ...
+cr2custom(differential) = ...
+```
+Secondly, we use the `test_rrule` function to test the gradients
+```julia
+test_rrule(complicated, 2.3, 6.1; rrule_f=ad_rrule)
+```
+by specifying the `ad_rrule` as the `rrule_f` keyword argument.
+
 ## Custom finite differencing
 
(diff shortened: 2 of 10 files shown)


oxinabox:
We should think a bout the relationship between this and https://github.com/JuliaDiff/ChainRulesCore.jl/pull/363/
In particular that `frule_f` and `rrule_f` 
have basically (literally?) the same requirements as the `frule_via_ad` and `rrule_via_ad` from that PR.
and so when using this for your AD you are probably having both defined.

src/check_result.jl
```diff
@@ -30,8 +30,8 @@ for (T1, T2) in ((AbstractThunk, Any), (AbstractThunk, AbstractThunk), (Any, Abs
     end
 end
 
-test_approx(::ZeroTangent, x, msg=""; kwargs...) = test_approx(zero(x), x, msg; kwargs...)
-test_approx(x, ::ZeroTangent, msg=""; kwargs...) = test_approx(x, zero(x), msg; kwargs...)
+test_approx(::AbstractZero, x, msg=""; kwargs...) = test_approx(zero(x), x, msg; kwargs...)
+test_approx(x, ::AbstractZero, msg=""; kwargs...) = test_approx(x, zero(x), msg; kwargs...)
```
mzgubic, line 34: The reason I've changed this is somewhat convoluted and the dependency goes the wrong way. But hear me out:

When writing a helper `zygote_rrule` function in Zygote (see https://github.com/FluxML/Zygote.jl/pull/987/) one needs to transform `nothing` (representing a zero gradient) to either `ZeroTangent` or `NoTangent`.
If we decide to transform to `ZeroTangent`, `test_rrule` complains that `NoTangent` should be used for non-perturbable arguments (say indices).
If we decide to transform to `NoTangent`, the `test_approx` function throws a fit, since `test_approx(::NoTangent, x)` is not defined and it falls back to the default implementation.

Alternatives:
- Do not provide `zygote_rrule` function in Zygote. It would still be possible to write a conversion for each separate function (where we know which `AbstractZero` the `nothing` represents). This seems to be the cleanest option but on the other hand it would greatly reduce the usability of `rrule_f` kwarg for testing Zygote gradients, since one would need to write a custom rrule wrapper for each function that is being tested. Eww.
- Pirate `test_approx(::NoTangent, x)` in Zygote. (probably very bad)

All of the above aside, I am leaning towards the current state where `test_approx(::NoTangent, x)` is not defined. However in this case I suspect it would be better to be practical. After all, we are happy to define `+(NoTangent, x)` to be equal to `x`, so for all intents and purposes we are implicitly saying that `test_approx(::NoTangent, x) = test_approx(zero(x), x)`.

test/testers.jl
```diff
@@ -530,4 +530,29 @@ end
...
+        only2x(x, y) = 2x
+        custom(::typeof(only2x), x, y) = only2x(x, y), Δ -> (NoTangent(), 2Δ, ZeroTangent())
+        wrong1(::typeof(only2x), x, y) = only2x(x, y), Δ -> (ZeroTangent(), 2Δ, ZeroTangent())
+        wrong2(::typeof(only2x), x, y) = only2x(x, y), Δ -> (NoTangent(), 2.1Δ, ZeroTangent())
+        wrong3(::typeof(only2x), x, y) = only2x(x, y), Δ -> (NoTangent(), 2Δ)
+
+        test_rrule(only2x, 2.0, 3.0; rrule_f=custom, check_inferred=false)
+        @test fails(() -> test_rrule(only2x, 2.0, 3.0; rrule_f=wrong1, check_inferred=false))
+        @test fails(() -> test_rrule(only2x, 2.0, 3.0; rrule_f=wrong2, check_inferred=false))
+        @test_broken fails(() -> test_rrule(only2x, 2.0, 3.0; rrule_f=wrong3, check_inferred=false))
```
mzgubic, line 544:
```suggestion
        @test fails(() -> test_rrule(only2x, 2.0, 3.0; rrule_f=wrong3, check_inferred=false))
```
^ This is what it should be, really, but I can't tell why the `fails` doesn't work as expected in this case. (`test_rrule` actually does fail here, as it should, but `fails` does not catch that). Perhaps something to do with the recent `ExprAndMsg` hack @oxinabox?

src/check_result.jl
oxinabox: Rather than allowing it in general between `AbstractZero` and anything, which would allow between `NoTangent()` and `0.0` which seems bad.
would it be pragmatic to allow it only between `AbstractZero` and `NoTangent` ?


> Pirate test_approx(::NoTangent, x) in Zygote. (probably very bad)

Doing this in Zygote's tests though doesn't seem that awful.
(With appropriate comments)

test/testers.jl
oxinabox: I will investigate

src/check_result.jl
mzgubic: Oh, it is exactly the `NoTangent()` and `0.0` that we need, sorry that wasn't clear

test/testers.jl
oxinabox: This is a bug in ChainRules
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/167

I will fix tomorrow

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.7.0"
+version = "0.7.1"
```
mzgubic, line 3:
```suggestion
version = "0.7.3"
```

src/check_result.jl
mzgubic: > We should think a bout the relationship between this and [JuliaDiff/ChainRulesCore.jl#363](https://github.com/JuliaDiff/ChainRulesCore.jl/pull/363)
> In particular that `frule_f` and `rrule_f`
> have basically (literally?) the same requirements as the `frule_via_ad` and `rrule_via_ad` from that PR.
> and so when using this for your AD you are probably having both defined.

Yeah that is a good point, I can't think of any differences on the top of my head. One consideration perhaps is how to deal with kwargs, since Zygote pretty much ignores them.

mzgubic:
>We should think a bout the relationship between this and JuliaDiff/ChainRulesCore.jl#363
In particular that frule_f and rrule_f
have basically (literally?) the same requirements as the frule_via_ad and rrule_via_ad from that PR.
and so when using this for your AD you are probably having both defined.

Yeah that is a good point, I can't think of any differences on the top of my head. One consideration perhaps is how to deal with kwargs, since Zygote pretty much ignores them.

oxinabox:
> One consideration perhaps is how to deal with kwargs, since Zygote pretty much ignores them.

So does rrule and frule.
I have an idea how to make them work, that works for Zygote and rrule/frule: defining it at the positional kwsorter level.
But we don't need to worry about it til later.
It's so far never come up as a problem.

mzgubic:
> > One consideration perhaps is how to deal with kwargs, since Zygote pretty much ignores them.
> 
> So does rrule and frule.
> I have an idea how to make them work, that works for Zygote and rrule/frule: defining it at the positional kwsorter level.
> But we don't need to worry about it til later.
> It's so far never come up as a problem.
> In the years of people doing AD in julia

The differences is that in `f/rrules` one can pass the kwargs to the primal pass, but in Zygote
```julia
pullback((x, y) -> cat(x, y; dims=1), rand(3), rand(4))
# pullback(cat, rand(3), rand(4); dims=1) # does not work
```

oxinabox:
>The differences is that in f/rrules one can pass the kwargs to the primal pass, but in Zygote
>```
>pullback((x, y) -> cat(x, y; dims=1), rand(3), rand(4))
># pullback(cat, rand(3), rand(4); dims=1) # does not work
>```

True. So yes Zygote's `rrule_f` and it's `rrule_via_ad` would need to accept kwargs, then put them into a closure which it calls it's `pullback` on  (maan I hate that name).

Lets hold off merging this PR til we are about to merge the calling back into AD one.

Just on the principle of delaying decisions til the last minute so that we have maximum information

src/check_result.jl
oxinabox: Where do the `NoTangents` from Zygote come from?
`nothing` becomes `ZeroTangent`.
Which should allos the comparison

https://github.com/FluxML/Zygote.jl/blob/12f5c1d75eeaa8c7a818f2db7f8d082956c00cac/src/compiler/chainrules.jl#L62

oxinabox: hmm Ok.

mzgubic: Yeah, thanks, it's already changed in the code

docs/src/index.md
```diff
@@ -145,6 +145,67 @@ In particular, when specifying the input tangents to [`test_frule`](@ref) and th
 As these tangents are used to seed the derivative computation.
 Inserting inappropriate zeros can thus hide errors.
 
+## Testing higher order functions
+
+Higher order functions, such as `map`, take a function (or a functor) `f` as an argument.
+`f/rrule`s for these functions call back into AD to compute the `f/rrule` of `f`.
+To test these functions, we use a dummy AD system, which simply calls the appropriate rule for `f` directly.
+For this reason, when testing `map(f, collection)`, the rules for `f` need to be defined.
+The `RuleConfig` for this dummy AD system is the default one, and does not need to be provided.
```
oxinabox, line 154: might be worth hard-coding a link to the ChainRulesCore docs

oxinabox (approved):

oxinabox:
Julia 1.0 error is real
LoadError: error compiling top-level scope: type definition not allowed inside a local scope

mzgubic:
yeah on it

PR #168, opened 2021-06-04
Test the number of outputs in `frule` and `rrule` are correct

### LINKED ISSUE #167: test_rrule doesn't catch returning wrong number of outputs
Consider a primal function with 2 inputs.
It's pullback should thus have 3 outputs (one for each input, plus one for the function itself).

`test_rrule` doesn't notice the following mistake

```julia
using ChainRulesTestUtils
using ChainRulesCore

foo(x, y) = x + 2y

function ChainRulesCore.rrule(::typeof(foo), x, y)
    foo_pullback(dz) = NoTangent(), dz

    # Should be
    #foo_pullback(dz) = NoTangent(), dz, 2dz

    return foo(x,y), foo_pullback
end

test_rrule(foo, 21.0, 32.0)
```

This is because `zip` in julia truncates to the shortest.
So as long as the ones that it does have are right it passes it.

The mistake is in:
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/1f507669ee4780c188d69c01d6fde7f2db3fa672/src/testers.jl#L207

We just need to add some extra tests before we get to that point

mzgubic:
Closes #167

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.7.2"
+version = "0.7.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.7.1"
+version = "0.7.2"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -111,6 +111,7 @@ function test_frule(
         res = frule((NoTangent(), deepcopy(ẋs)...), f, deepcopy(xs)...; deepcopy(fkwargs)...)
         res === nothing && throw(MethodError(frule, typeof((f, xs...))))
         res isa Tuple || error("The frule should return (y, ∂y), not $res.")
+        @test_msg "frule should return (y, ∂y), not $res" length(res) == 2
         Ω_ad, dΩ_ad = res
         Ω = f(deepcopy(xs)...; deepcopy(fkwargs)...)
         test_approx(Ω_ad, Ω; isapprox_kwargs...)
@@ -191,6 +192,9 @@ function test_rrule(
         ∂self = ∂s[1]
         x̄s_ad = ∂s[2:end]
         @test ∂self === NoTangent()  # No internal fields
+        msg = "The pullback must return `(∂self, ∂args...)`, where `length(∂args) == " *
+            "length(args)`, and `args` are the arguments of the primal function."
+        @test_msg msg length(x̄s_ad) == length(inputs)
 
         # Correctness testing via finite differencing.
         # TODO: remove Nothing when https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/113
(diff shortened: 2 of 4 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "0.7.1"
+version = "0.7.2"
```
mzgubic, line 3:
```suggestion
version = "0.7.1"
```

src/testers.jl
```diff
@@ -110,7 +110,8 @@ function test_frule(
         end
         res = frule((NoTangent(), deepcopy(ẋs)...), f, deepcopy(xs)...; deepcopy(fkwargs)...)
         res === nothing && throw(MethodError(frule, typeof((f, xs...))))
-        res isa Tuple || error("The frule should return (y, ∂y), not $res.")
+        msg = "The frule should return (y, ∂y), not $res."
+        @test_msg msg res isa Tuple && length(res) == 2
```
oxinabox, line 114: Could write this as


```suggestion
        @test_msg msg res isa Tuple{Any,Any}
```

What do you think?

src/testers.jl
```diff
@@ -191,6 +192,9 @@ function test_rrule(
         ∂self = ∂s[1]
         x̄s_ad = ∂s[2:end]
         @test ∂self === NoTangent()  # No internal fields
+        msg = "The pullback must return `(∂self, ∂args...)`, where `length(∂args) == " *
+            "length(args)`, and `args` are the arguments of the primal function."
```
oxinabox, line 196: Maybe more direct about the problem
```suggestion
            "The pullback should return 1 cotangent for each primal input."
```

src/testers.jl
```diff
@@ -191,6 +192,9 @@ function test_rrule(
...
+        @test_msg msg length(x̄s_ad) == length(inputs)
```
oxinabox, line 197: Maybe we should rename these to be nicer in failure message? Or maybe they are good enough

test/testers.jl
```diff
@@ -495,6 +495,24 @@ end
...
+            foo(x, y) = x + 2y
+
+            function ChainRulesCore.frule((_, ẋ, ẏ), ::typeof(foo), x, y)
+                return foo(x, y), ẋ + 2ẏ, NoTangent() # extra derivative
+                #return foo(x, y), ẋ + 2ẏ # correct expression
+            end
+
+            function ChainRulesCore.rrule(::typeof(foo), x, y)
+                foo_pullback(dz) = NoTangent(), dz # missing derivative
+                #foo_pullback(dz) = NoTangent(), dz, 2dz # correct expression
```
oxinabox, line 509:
```suggestion
```

oxinabox (approved):

src/testers.jl
mzgubic, line 196:
```suggestion
        msg = "The pullback should return 1 cotangent for each primal input."
```

mzgubic:
I've stupidly merged this before the Integration test finished. Turns out that some ChainRules tests get broken by this, so I've reverted the changes and tagged another patch version of ChainRulesTestUtils.

Essentially what happens inside ChainRules is that there are some rrules for functions with optional arguments, e.g.
```julia
function rrule(
    ::typeof(pinv),
    x::AbstractVector{T},
    tol::Real = 0,
) where {T<:Union{Real,Complex}}
    y = pinv(x, tol)
    function pinv_pullback(Δy)
        ∂x = sum(abs2, parent(y)) .* vec(Δy') .- 2real(y * Δy') .* parent(y)
        return (NoTangent(), ∂x, ZeroTangent())
    end
    return y, pinv_pullback
end
```
Here, even if `pinv` is called with one argument `x`, `ZeroTangent()` is returned for `tol`. The other failure is similar.

PR #170, opened 2021-06-08
Make work on functors

### LINKED ISSUE #117: Make work on Functors
right now all the tests check they are not funning on functors.

With the new API in https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/116
we can have `f ⟂ ḟ` and `f ⟂ f̄``, which would be fine API-wise.

mzgubic:
Closes #117 

To do:
- [x] merge and tag [FiniteDifferences PR](https://github.com/JuliaDiff/FiniteDifferences.jl/pull/172), add compat. Needed to generate `rand_tangent` for `f` when testing constructors, i.e. `rand_tangent(Foo)`
- [x] implement `test_frule` for functors
- [x] implement `test_scalar` for functors
- [x] remove `_ensure_not_running_on_functor`

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -18,7 +18,6 @@ function test_scalar(f, z; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(),
     isapprox_kwargs = (; rtol=rtol, atol=atol, kwargs...)
 
     @testset "test_scalar: $f at $z" begin
-        _ensure_not_running_on_functor(f, "test_scalar")
         # z = x + im * y
         # Ω = u(x, y) + im * v(x, y)
         Ω = f(z; fkwargs...)
@@ -30,8 +29,9 @@ function test_scalar(f, z; rtol=1e-9, atol=1e-9, fdm=_fdm, fkwargs=NamedTuple(),
             test_frule(f, z ⊢ Δx; rule_test_kwargs...)
             if z isa Complex
                 # check that same tangent is produced for tangent 1.0 and 1.0 + 0.0im
-                _, real_tangent = frule((ZeroTangent(), real(Δx)), f, z; fkwargs...)
-                _, embedded_tangent = frule((ZeroTangent(), Δx), f, z; fkwargs...)
+                ḟ = rand_tangent(f)
+                _, real_tangent = frule((ḟ, real(Δx)), f, z; fkwargs...)
+                _, embedded_tangent = frule((ḟ, Δx), f, z; fkwargs...)
                 test_approx(real_tangent, embedded_tangent; isapprox_kwargs...)
             end
         end
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -100,15 +100,13 @@ function test_frule(
     isapprox_kwargs = (; rtol=rtol, atol=atol, kwargs...)
 
     @testset "test_frule: $f on $(_string_typeof(args))" begin
-        _ensure_not_running_on_functor(f, "test_frule")
-
         xẋs = auto_primal_and_tangent.(args)
         xs = primal.(xẋs)
-        ẋs = tangent.(xẋs)
+        tangents = (rand_tangent(f), tangent.(xẋs)...)
         if check_inferred && _is_inferrable(f, deepcopy(xs)...; deepcopy(fkwargs)...)
-            _test_inferred(frule, (NoTangent(), deepcopy(ẋs)...), f, deepcopy(xs)...; deepcopy(fkwargs)...)
+            _test_inferred(frule, (deepcopy(tangents)...), f, deepcopy(xs)...; deepcopy(fkwargs)...)
         end
-        res = frule((NoTangent(), deepcopy(ẋs)...), f, deepcopy(xs)...; deepcopy(fkwargs)...)
+        res = frule(deepcopy(tangents), f, deepcopy(xs)...; deepcopy(fkwargs)...)
         res === nothing && throw(MethodError(frule, typeof((f, xs...))))
         @test_msg "The frule should return (y, ∂y), not $res." res isa Tuple{Any,Any}
         Ω_ad, dΩ_ad = res
@@ -116,8 +114,8 @@ function test_frule(
         test_approx(Ω_ad, Ω; isapprox_kwargs...)
 
         # TODO: remove Nothing when https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/113
-        ẋs_is_ignored = isa.(ẋs, Union{Nothing,NoTangent})
-        if any(ẋs .== nothing)
+        is_ignored = isa.(tangents, Union{Nothing,NoTangent})
+        if any(tangents .== nothing)
             Base.depwarn(
                 "test_frule(f, k ⊢ nothing) is deprecated, use " *
                 "test_frule(f, k ⊢ NoTangent()) instead for non-differentiable ks",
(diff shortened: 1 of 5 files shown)


src/testers.jl
```diff
@@ -186,44 +185,42 @@ function test_rrule(
...
                 :test_rrule,
             )
         end
 
-        x̄s_fd = _make_j′vp_call(fdm, (xs...) -> f(xs...; fkwargs...), ȳ, xs, x̄s_is_dne)
-        for (accumulated_x̄, x̄_ad, x̄_fd) in zip(accumulated_x̄, x̄s_ad, x̄s_fd)
-            if accumulated_x̄ isa Union{Nothing,NoTangent}  # then we marked this argument as not differentiable # TODO remove once #113
-                @assert x̄_fd === nothing  # this is how `_make_j′vp_call` works
-                x̄_ad isa ZeroTangent && error(
+        fd_cotangents = _make_j′vp_call(fdm, (f, xs...) -> f(xs...; fkwargs...), ȳ, (f, xs...), is_ignored)
```
mzgubic, line 204: This is the main change, i.e. from
`(xs...) -> f(xs...; fkwargs...)` to
`(f, xs...) -> f(xs...; fkwargs...)`
making the `f` an argument. The rest of the changes are adding an extra cotangent for `f` via `rand_tangent`, and renaming the variables from `x̄_something` to `something_cotangent` since it also includes `f̄`

src/testers.jl
```diff
@@ -100,24 +100,22 @@ function test_frule(
     isapprox_kwargs = (; rtol=rtol, atol=atol, kwargs...)
 
     @testset "test_frule: $f on $(_string_typeof(args))" begin
-        _ensure_not_running_on_functor(f, "test_frule")
-
         xẋs = auto_primal_and_tangent.(args)
         xs = primal.(xẋs)
-        ẋs = tangent.(xẋs)
+        tangents = (rand_tangent(f), tangent.(xẋs)...)
```
oxinabox, line 105: We should make  `f` also participage in `auto_primal_and_tangents`
and for the functors one would optionally write: `test_frule(f ⊢  ḟ, ...)`

oxinabox: Yes, we do.
KeyedArrays use call overloading for accessing by key rather than by integer index.
And for those we likely want to use another KeyedArray as the cotangent

mzgubic: Done, and ready for review

src/testers.jl
```diff
@@ -100,17 +100,20 @@ function test_frule(
...
 
     @testset "test_frule: $f on $(_string_typeof(args))" begin
-        xẋs = auto_primal_and_tangent.(args)
-        xs = primal.(xẋs)
-        tangents = (rand_tangent(f), tangent.(xẋs)...)
-        if check_inferred && _is_inferrable(f, deepcopy(xs)...; deepcopy(fkwargs)...)
-            _test_inferred(frule, (deepcopy(tangents)...,), f, deepcopy(xs)...; deepcopy(fkwargs)...)
+
+        primals_and_tangents = auto_primal_and_tangent.((f, args...))
+        func, xs = Iterators.peel(primal.(primals_and_tangents))
```
oxinabox, line 105: It seems like most places (except line 116) that we use this, we could be using 
```suggestion
        primals = primal.(primals_and_tangents)
```
and for line 116 we might want to define a helper closure locally
`call_on_copy(f, xs...) = deepcopy(f)(deepcopy(xs)...; deepcopy(fkwargs)...)`
Which we also need for line 132.

src/testers.jl
```diff
@@ -100,17 +100,20 @@ function test_frule(
...
+        tangents = tangent.(primals_and_tangents)
+
+        if check_inferred && _is_inferrable(func, deepcopy(xs)...; deepcopy(fkwargs)...)
```
oxinabox, line 108: note that we need to deep copy `func` here also.
 as well as on line 109, 112, and 116
 
 Possibly this also could use `_is_inferable(call_on_copy(primals...))` ?

src/testers.jl
```diff
@@ -220,7 +224,7 @@ function test_rrule(
             if accum_cotangent isa Union{Nothing,NoTangent}  # then we marked this argument as not differentiable # TODO remove once #113
                 @assert fd_cotangent === nothing  # this is how `_make_j′vp_call` works
                 ad_cotangent isa ZeroTangent && error(
-                    "The pullback in the rrule for $f function should use NoTangent()" *
+                    "The pullback in the rrule for $func function should use NoTangent()" *
```
oxinabox, line 227: possibly we don't need to mention the func here.
It should be in the stack trace anyway

src/testers.jl
mzgubic: Not sure how this would work? Otherwise it's ready for re-review

mzgubic:
No clue why the Documentation test fails, and I can't reproduce it locally even though the Julia version, Documenter version, and ChainRulesCore/TestUtils versions are all the same.

src/testers.jl
```diff
@@ -129,9 +132,9 @@ function test_frule(
         # Correctness testing via finite differencing.
         dΩ_fd = _make_jvp_call(
             fdm,
-            (f, xs...) -> f(deepcopy(xs)...; deepcopy(fkwargs)...),
+            (f, xs...) -> call_on_copy(f, xs...),
```
oxinabox, line 135: Isn't this just the same as:
```suggestion
            call_on_copy,
```

src/testers.jl
```diff
@@ -214,7 +218,7 @@ function test_rrule(
             fdm,
             (f, xs...) -> f(xs...; fkwargs...),
```
oxinabox, line 219: should we do the same trick here, only insted of `call_on_copy` it is just `call`?
Which we would use here and on line 195?

src/testers.jl
oxinabox: Ah year it wouldn't quite work

oxinabox:
> No clue why the Documentation test fails, and I can't reproduce it locally even though the Julia version, Documenter version, and ChainRulesCore/TestUtils versions are all the same.

Hack to solve this.
Lets install the GHA that makes suggestions to fix doc-tests.
We can copy it from: https://github.com/beacon-biosignals/KeywordSearch.jl/pull/18

We kinda want to do that anyway.




If you need to get this merged can change that bit of the docs to be a  
````
```julia
````
block rather than a 
````
```@doctest
````
block.

Else can make another PR to do that stuff, and then rebase this on top of it.

Your call

mzgubic:
> > No clue why the Documentation test fails, and I can't reproduce it locally even though the Julia version, Documenter version, and ChainRulesCore/TestUtils versions are all the same.
> 
> Hack to solve this.
> Lets install the GHA that makes suggestions to fix doc-tests.
> We can copy it from: [beacon-biosignals/KeywordSearch.jl#18](https://github.com/beacon-biosignals/KeywordSearch.jl/pull/18)
> 
> We kinda want to do that anyway.
> 
> If you need to get this merged can change that bit of the docs to be a
> 
> ```
> ```julia
> ```
> 
> block rather than a
> 
> ```
> ```@doctest
> ```
> 
> block.
> 
> Else can make another PR to do that stuff, and then rebase this on top of it.
> 
> Your call

The problem is that I can't find the place where I should make the change.😬 The doctest that is meant to be failing does not exist explicitly in the code anywhere. It looks like it was generated from `@autodocs` but the current docs don't show it either. It's a ghost.

The failure of the doctest is that the stacktrace is coming from a different line. So it looks like it is picking up some doctest that is itself using a different version of ChainRulesTestUtils (probably master?). I wonder whether we can just try merging an see if it works?

simeonschaub:
Oh, sorry, looks like that's the doctest I introduced in #165 that's failing. I wouldn't recommend merging this before manually merging master into this branch. The failure is most likely real, because I couldn't figure out how to not make the error output I wanted to show not print the location of the testset in `test_rrule`. You could of course just correct the line number, but I agree that that's not a great solution, so I'm fine with not making that a doctest too.

nickrobinson251:
Alternatively could you Documenters Filter functionality to ignore the line number info

mzgubic:
> Oh, sorry, looks like that's the doctest I introduced in #165 that's failing. I wouldn't recommend merging this before manually merging master into this branch. The failure is most likely real, because I couldn't figure out how to not make the error output I wanted to show not print the location of the testset in `test_rrule`. You could of course just correct the line number, but I agree that that's not a great solution, so I'm fine with not making that a doctest too.

Oh, thanks! This was driving me crazy. I am still confused though - I thought CI runs on my branch? I haven't merged your changes. So it appears at least the Documenter action is running on what master would look like if this would have been merged? How does it deal with conflicts? 😬 How d...wh...? !??!? 😂

oxinabox (approved):

PR #171, opened 2021-06-08
add support for two-arg `@maybe_inferred`

simeonschaub:
Also adds some direct tests for `@maybe_inferred`.

### DIFF
--- a/test/testers.jl
+++ b/test/testers.jl
@@ -551,7 +551,7 @@ end
         f_noninferrable(x) = Ref{Real}(x)[]
 
         @test @maybe_inferred(identity(1)) == 1
-        errors("not inferrable") do
+       @test errors("not inferrable") do
             @maybe_inferred f_noninferrable(1)
         end
         @test @maybe_inferred(Real, f_noninferrable(1)) == 1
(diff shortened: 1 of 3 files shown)


test/testers.jl
```diff
@@ -546,4 +546,22 @@ end
         test_frule(f_notimplemented, randn(), randn())
         test_rrule(f_notimplemented, randn(), randn())
     end
+
+    @testset "@maybe_inferred" begin
+        f_noninferrable(x) = Ref{Real}(x)[]
+
+        @test @maybe_inferred(identity(1)) == 1
+        errors("not inferrable") do
```
oxinabox, line 554: `errors` returns `true` if the error occured.
```suggestion
       @test errors("not inferrable") do
```

simeonschaub: Ah, makes sense!

test/testers.jl
```diff
@@ -551,7 +551,7 @@ end
         f_noninferrable(x) = Ref{Real}(x)[]
 
         @test @maybe_inferred(identity(1)) == 1
-        errors("not inferrable") do
+       @test errors("not inferrable") do
```
oxinabox: ooops this is prob me not counting my indenting right
```suggestion
        @test errors("not inferrable") do
```

oxinabox (approved):

PR #175, opened 2021-06-11
do not check for thunks

mzgubic:
This was there to prevent overthunking, but fails for some cases where thunking is useful. The instructions on when to thunk are in the docs


oxinabox:
I believe @mcabbott  also wanted this one removed?
It is one less place where the tests can be persnicketty.

oxinabox (approved):
I am surprised we didn't have tests for this

mcabbott:
Yes this was #144, still seems worth removing IMO.

Issue #159, opened 2021-06-01
Test all kinds of tangent types (Thunk, ZeroTangent, Tangent{T} etc)


oxinabox:
This is a generalization of #98  and is a key part of https://github.com/JuliaDiff/ChainRules.jl/issues/408

Lets say we have a tanget `dx`
  we should also test that in it's place we can put `@thunk(x̄)`, as well as `ZeroTangent`.

Further, following https://github.com/JuliaDiff/ChainRulesCore.jl/issues/286
if `dx <: AbstractArray` (other than `Array`) we should test `canonicalize(Composite{typeof(x)}, dx)`
and conversely
if `dx <: Composite{P} where P<:AbstractArray` we should test `canonicalize(P, dx)`

mzgubic:
Just to confirm: for `test_rrule` we just want to change the `output_tangent`, and not the cotangents of the `args`, right?

And conversely, for `test_frule`, we just want the tangents of the `args`, and not the tangent of the output?

mzgubic:
Also, the current implementation I have in mind is something like
```julia
function test_rrule(f, args; output_tangent=Auto(), test_other_tangents=true; other_kwargs...)
    y = func(xs...; fkwargs...)
    ȳ = output_tangent isa Auto ? rand_tangent(y) : output_tangent
    if test_other_tangents
        for ybar in _all_possible_tangents(ȳ)
            test_rrule(f, args; output_tangent=ybar, test_other_tangents=false, other_kwargs...)
        end
    else
        do_the_current_thing()
    end
end
    
```
It's kind of ugly, do we have other ideas?

oxinabox:
> Just to confirm: for test_rrule we just want to change the output_tangent, and not the cotangents of the args, right?
> And conversely, for test_frule, we just want to change the tangents of the args, and not the tangent of the output?

This is correct

> One other point is that requiring new output_tangent to pass the tests is a breaking change. I was initially planning to do this in two parts, i.e. adding ZeroTangents and thunks first, and adding canonicalize once we figure out the arrays.

Two parts sounds better.
Yes it means multiple rounds of breaking changes, but it is ChainRulesTestUtils so it is only a test dependency.

Potentially we can do something like what we have like we do for inference and have it's default controlled by a global?
That way we can start to roll it out for ChainRules.jl then once we have it all working and are happy with it we can change the default and tag a breaking release.

Possibly rather than a bool we want a typed object, for extensibility.
[...]

PR #176, opened 2021-06-11
Test other tangents

mzgubic:
First part of https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/159

Lighter alternative to https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/174, this just runs the pullback to check there is no errors

### DIFF
--- a/src/ChainRulesTestUtils.jl
+++ b/src/ChainRulesTestUtils.jl
@@ -13,6 +13,7 @@ import FiniteDifferences: rand_tangent
 
 const _fdm = central_fdm(5, 1; max_range=1e-2)
 const TEST_INFERRED = Ref(true)
+const TRANSFORMS_TO_ALT_TEST_TANGENTS = Ref{Vector{Function}}([])
 
 export TestIterator
 export test_approx, test_scalar, test_frule, test_rrule, generate_well_conditioned_matrix
--- a/src/ChainRulesTestUtils.jl
+++ b/src/ChainRulesTestUtils.jl
@@ -13,7 +13,7 @@ import FiniteDifferences: rand_tangent
 
 const _fdm = central_fdm(5, 1; max_range=1e-2)
 const TEST_INFERRED = Ref(true)
-const TRANSFORMS_TO_ALT_TEST_TANGENTS = Ref{Vector{Function}}([])
+const TRANSFORMS_TO_ALT_TANGENTS = Function[] # e.g. [x -> @thunk(x), _ -> ZeroTangent(), x -> rebasis(x)]
 
 export TestIterator
 export test_approx, test_scalar, test_frule, test_rrule, generate_well_conditioned_matrix
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -80,7 +80,7 @@ end
 # Keyword Arguments
    - `output_tangent` tangent to test accumulation of derivatives against
      should be a differential for the output of `f`. Is set automatically if not provided.
-   - `tangent_transforms=TRANSFORMS_TO_ALT_TEST_TANGENTS[]`: a vector of functions that
+   - `tangent_transforms=TRANSFORMS_TO_ALT_TANGENTS`: a vector of functions that
       transform the passed argument tangents into multiple tangents that should be tested.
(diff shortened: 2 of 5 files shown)


src/ChainRulesTestUtils.jl
```diff
@@ -13,6 +13,7 @@ import FiniteDifferences: rand_tangent
 
 const _fdm = central_fdm(5, 1; max_range=1e-2)
 const TEST_INFERRED = Ref(true)
+const TRANSFORMS_TO_ALT_TEST_TANGENTS = Ref{Vector{Function}}([])
```
oxinabox, line 16: Why is the `Ref` here?
we should be ok with just `Vector{Function}` since if we want to reset it we can use `empty!` and `append!`/`push!` etc

oxinabox: Also shouldn't it be `TRANSFORMS_TO_TEST_ALT_TANGENTS`  ?
we are testing alternatative tangents, not alternatively testing tangents.

src/testers.jl
```diff
@@ -80,6 +80,10 @@ end
 # Keyword Arguments
    - `output_tangent` tangent to test accumulation of derivatives against
      should be a differential for the output of `f`. Is set automatically if not provided.
+   - `tangent_transforms=TRANSFORMS_TO_ALT_TEST_TANGENTS[]`: a vector of functions that
+      transform the passed argument tangents into multiple tangents that should be tested.
+      Note that the alternative tangents are only passed through the frule, and are not
```
oxinabox, line 85: maybe a comment here to say that it asserts they don't error.
And somehting about if you need to test their correctness use a seperate `test_frule` with `⊢` and `output_tangent`, passing `tangent_transforms=[]`

src/testers.jl
```diff
@@ -122,7 +127,12 @@ function test_frule(
             _test_inferred(frule_f, deepcopy(config), deepcopy(tangents), deepcopy(primals)...; deepcopy(fkwargs)...)
         end
 
-        res = frule_f(deepcopy(config), deepcopy(tangents), deepcopy(primals)...; deepcopy(fkwargs)...)
+        # test that rules work for other tangents
+        for tangent_transform in tangent_transforms
+            call_on_copy(frule_f, config, tangent_transform.(tangents), primals...)
```
oxinabox, line 132: maybe
```suggestion
            @test_msg "Shouldn't error on $tangent_transform tangents" call_on_copy(frule_f, config, tangent_transform.(tangents), primals...) isa Any
```
though idk if `@test_msg` works right on things that throw errors rather than failing

src/testers.jl
```diff
@@ -122,7 +127,12 @@ function test_frule(
             _test_inferred(frule_f, deepcopy(config), deepcopy(tangents), deepcopy(primals)...; deepcopy(fkwargs)...)
         end
 
-        res = frule_f(deepcopy(config), deepcopy(tangents), deepcopy(primals)...; deepcopy(fkwargs)...)
+        # test that rules work for other tangents
+        for tangent_transform in tangent_transforms
```
oxinabox, line 131: maybe break this out in to a seperate function?

oxinabox (approved):
many comments but I am sure you can address them without me needing to rereview.
all my comments about frule apply to rrule as well.

I guess it is out of scope to teest `add!!` that can be another PR.

Do we want to also give some examples, suich as `_->ZeroTangent(x)`  `x->@thunk(x)` ?
Or even some predefined ones like those?

Should the tranforms accept 2 args? a primal and tangent, so we can generate `p,t->zero(p)` ?
I think probably not, because we actually have a fixed set we want to use in the longer term, and none of those need it
but it is worth thingking about.

src/ChainRulesTestUtils.jl
mzgubic: Oh, interesting, I was using 'test' as a noun, not a verb. I.e. these are the 'transforms to alternative test tangents'. perhaps we can just leave out `TEST_`

mzgubic:
Nice catch, I've added `add!!` tests, will leave it open until tomorrow in case you wanted another look

PR #177, opened 2021-06-15
fix mistake in non-tuple output message

oxinabox:
`∂s` was renamed to `ad_cotangents`, but we missed it in this error message.
So I fixed that and I also changed it to use `test_msg` rather than erroring.
added a test for this case, making sure it failed rather than erroring.

I also changed the length one after to not define a variable called `msg` since that was confusing as to which was meant.

### DIFF
--- a/.github/workflows/fix_doctests.yml
+++ b/.github/workflows/fix_doctests.yml
@@ -14,12 +14,12 @@ jobs:
           version: ${{ matrix.julia-version }}
       - uses: actions/checkout@v1
       - name: Fix doctests
-	    shell: julia --project=docs/ {0}
-		run: |
-			using Pkg:Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
-			using Documenter
-			using ChainRulesTestUtils
-			doctest(ChainRulesTestUtils, fix=true)
+        shell: julia --project=docs/ {0}
+        run: |
+          using Pkg:Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()
+          using Documenter
+          using ChainRulesTestUtils
+          doctest(ChainRulesTestUtils, fix=true)
       - uses: reviewdog/action-suggester@v1
         if: github.event_name == 'pull_request'
         with:
--- a/.github/workflows/fix_doctests.yml
+++ b/.github/workflows/fix_doctests.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Fix doctests
         shell: julia --project=docs/ {0}
         run: |
-          using Pkg:Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()
+          using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()
           using Documenter
           using ChainRulesTestUtils
           doctest(ChainRulesTestUtils, fix=true)
--- a/.github/workflows/fix_doctests.yml
+++ b/.github/workflows/fix_doctests.yml
@@ -20,6 +20,8 @@ jobs:
           using Documenter
           using ChainRulesTestUtils
           doctest(ChainRulesTestUtils, fix=true)
+          # don't push changes to Manifest in suggestions, as it removes `path=..`
+          run(`git restore docs/Manifest.toml`)
       - uses: reviewdog/action-suggester@v1
         if: github.event_name == 'pull_request'
         with:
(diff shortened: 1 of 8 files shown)


src/testers.jl
```diff
@@ -214,9 +214,14 @@ function test_rrule(
...
-        ad_cotangents isa Tuple || error("The pullback must return (∂self, ∂args...), not $∂s.")
-        msg = "The pullback should return 1 cotangent for the primal and each primal input."
-        @test_msg msg length(ad_cotangents) == 1 + length(args)
+        @test_msg(
+            "The pullback must return a Tuple (∂self, ∂args...)",
+            ad_cotangents isa Tuple
+        )
+        @test_msg(
+            "The pullback should return 1 cotangent for the primal and each primal input.",
+            length(ad_cotangents) == 1 + length(args)
```
mzgubic, line 223:
```suggestion
            length(ad_cotangents) == length(primals)
```

mzgubic (approved):
Thanks, LGTM, just needs fixing the docs

.github/workflows/fix_doctests.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          version: ${{ matrix.julia-version }}
+      - uses: actions/checkout@v1
+      - name: Fix doctests
+	    shell: julia --project=docs/ {0}
+		run: |
+			using Pkg:Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
+			using Documenter
+			using ChainRulesTestUtils
+			doctest(ChainRulesTestUtils, fix=true)
```
simeonschaub, line 22: Looks like there might be something off here with tabs/spaces?

oxinabox: ah yeah, new computer, I haven't yet fixed all my editors

.github/workflows/fix_doctests.yml
```diff
@@ -0,0 +1,27 @@
...
+        julia-version: [1.6]
+    steps:
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: ${{ matrix.julia-version }}
+      - uses: actions/checkout@v1
+      - name: Fix doctests
+        shell: julia --project=docs/ {0}
+        run: |
+          using Pkg:Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()
```
oxinabox, line 19:
```suggestion
          using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()
```

.github/workflows/fix_doctests.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          version: ${{ matrix.julia-version }}
+      - uses: actions/checkout@v1
+      - name: Fix doctests
+        shell: julia --project=docs/ {0}
+        run: |
+          using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()
+          using Documenter
+          using ChainRulesTestUtils
+          doctest(ChainRulesTestUtils, fix=true)
```
simeonschaub, line 22: Might also want to use the same trick as in https://github.com/JuliaDiff/FiniteDifferences.jl/pull/173/commits/7077850ca5e92ec8caa459174fc3e7b2004c54a1 here

simeonschaub (approved):

docs/Manifest.toml
```diff
@@ -11,15 +11,15 @@ uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
...
 deps = ["Compat", "LinearAlgebra", "SparseArrays"]
-git-tree-sha1 = "d659e42240c2162300b321f05173cab5cc40a5ba"
+git-tree-sha1 = "dbc9aae1227cfddaa9d2552f3ecba5b641f6cce9"
 uuid = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
-version = "0.10.4"
+version = "0.10.5"
 
 [[ChainRulesTestUtils]]
 deps = ["ChainRulesCore", "Compat", "FiniteDifferences", "LinearAlgebra", "Random", "Test"]
 path = ".."
```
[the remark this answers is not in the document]
oxinabox: No, bad bot.

.github/workflows/fix_doctests.yml
oxinabox: maybe, but we have to roll that out in many places so I will wait til we do that generally everywhere

simeonschaub: Yeah, we can wait until we run into any issues.

oxinabox:
the auto suggesting of fixes seems to be be doing a bad.
I am tempted to leave it in and see if it comes up good after: it might be some how confused since I have messed with documenter so much.

PR #178, opened 2021-06-16
Shorten testset names

mzgubic:
Shortens the testset names for alternative tangents. Did not notice in the original PR since all the tests passed.


oxinabox (approved):

oxinabox:
I feel like we should use `@test_msg` instead though.
Nexting testsets is a known cause of pain

PR #180, opened 2021-06-18
Remove Deprecations. Replace tangent_transforms functionality

### LINKED ISSUE #113: Remove the use of `nothing` to designate non-differentiability in `rrule_test`

### LINKED ISSUE #164: remove deprecations for `isapprox` and `f/rrule_test`

mzgubic:
Closes #113, closes #164.

~Do we also want https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/179?~

Make sure thunks are supported in rules in
- [x] [SpecialFunctions](https://github.com/JuliaMath/SpecialFunctions.jl/pull/329) 
- [x] [BlockDiagonals](https://github.com/invenia/BlockDiagonals.jl/pull/73)
- [x] [ChainRules](https://github.com/JuliaDiff/ChainRules.jl/pull/449)


oxinabox:
> Do we also want #179?

No, probably not.

Why are we releasing this right now?
Or are we just getting ready with the PR open?

mzgubic:
Yeah, just making sure we don't forget to deprecate things this time. We are planning to release two breaking versions though, right? - One which requires rules to support thunks (0.8) and one which requires the support for ZeroTangent as well (1.0)?

oxinabox:
> We are planning to release two breaking versions though, right? - One which requires rules to support thunks (0.8) and one which requires the support for ZeroTangent as well (1.0)?

I was thinking just 1.

oxinabox:
looks like we have some issues with it calling `Float64(::Thunk)`
I will debug this later today.

oxinabox:
This is now passing tests.
Because I added the missing unthunks.
Many of them were for `frules`.
I don't think using thunks in `frules` is a good idea because it breaks mutation support.

I think we should stop testing thunks in frules.

mzgubic:
Very naive, but how does it break mutation support? Is it that frules which support mutation can't receive thunks?

oxinabox:
> Very naive q, but how does it break mutation support? Is it that frules which support mutation can't receive thunks?

frules for mutating primal functions can't recieve thunks.
The first rule of mutation support club, is that mutation of the primal needs to result in matching mutation of the tangent.
Thunks are effectively immutable.
In general, for this reason, mutation support is much more fussy about the types involved.

Shall i remove that from this PR?
Probably easiest as I can see more clearly which frules i changed that i should not have

mzgubic:
Thanks, makes sense, I was just wondering whether there was anything else to it.

I assume mutation support is more important, right? So let's remove the default? I can also do that

oxinabox:
>  I assume mutation support is more important, right? So let's remove the default? I can also do that

I will do this

test/testers.jl
```diff
@@ -33,12 +33,12 @@ Base.iterate(f::Foo, state) = iterate(f.a, state)
...
     foo = Foo(a)
     function Foo_pullback(Δfoo)
-        return NoTangent(), Δfoo.a
+        return NoTangent(), unthunk(Δfoo).a
     end
     return foo, Foo_pullback
 end
 function ChainRulesCore.frule((_, Δa), ::Type{Foo}, a)
-    return Foo(a), Foo(Δa)
+    return Foo(a), Foo(unthunk(Δa))
```
oxinabox, line 41:
```suggestion
    return Foo(a), Foo(Δa)
```

test/testers.jl
```diff
@@ -49,7 +49,8 @@ function ChainRulesCore.rrule(f::Foo, x)
     end
     return y, Foo_pullback
 end
-function ChainRulesCore.frule((Δf, Δx), f::Foo, x)
+function ChainRulesCore.frule((Δf_, Δx), f::Foo, x)
+    Δf = unthunk(Δf_)
```
oxinabox, line 53:
```suggestion
function ChainRulesCore.frule((Δf, Δx), f::Foo, x)
```

test/testers.jl
```diff
@@ -158,7 +159,7 @@ struct MySpecialConfig <: RuleConfig{Union{MySpecialTrait}} end
 
         @testset "check not inferred in frule" begin
             function ChainRulesCore.frule((_, Δx), ::typeof(f_noninferrable_frule), x)
-                return (x, x > 0 ? Float64(Δx) : Float32(Δx))
+                return (x, x > 0 ? Float64(unthunk(Δx)) : Float32(unthunk(Δx)))
```
oxinabox, line 162:
```suggestion
                return (x, x > 0 ? Float64(Δx) : Float32(Δx))
```

test/testers.jl
```diff
@@ -233,10 +234,13 @@ struct MySpecialConfig <: RuleConfig{Union{MySpecialTrait}} end
 
         @testset "check non-inferrable primal still passes if pullback inferrable" begin
             function ChainRulesCore.frule((_, Δx), ::typeof(f_inferrable_pullback_only), x)
-                return (x > 0 ? Float64(x) : Float32(x), x > 0 ? Float64(Δx) : Float32(Δx))
+                T  = x > 0 ? Float64 : Float32
+                return T(x), T(unthunk(Δx))
```
oxinabox, line 238:
```suggestion
                return T(x), T(Δx)
```

test/testers.jl
```diff
@@ -462,7 +468,8 @@ struct MySpecialConfig <: RuleConfig{Union{MySpecialTrait}} end
             return s
         end
 
-        function ChainRulesCore.frule((_, Δiter), ::typeof(iterfun), iter)
+        function ChainRulesCore.frule((_, Δiter_), ::typeof(iterfun), iter)
+            Δiter = unthunk(Δiter_)
```
oxinabox, line 472:
```suggestion
        function ChainRulesCore.frule((_, Δiter), ::typeof(iterfun), iter)
```

mzgubic:
should we test that we catch bad rules with this?

mzgubic:
I approve the commits added (but can't click approve). One thought - should we also keep the empty deprecated.jl file in the tests so that we don't forget to add it?

oxinabox:
>  One thought - should we also keep the empty deprecated.jl file in the tests so that we don't forget to add it?

Maybe, I think it is not a big deal either way, So I am not going to retrigger CI for that

docs/src/index.md
```diff
@@ -72,7 +72,7 @@ The call will test the `rrule` for function `f` at the point `x`, and similarly
 ```jldoctest ex
 julia> test_rrule(two2three, 3.33, -7.77);
 Test Summary:                            | Pass  Total
-test_rrule: two2three on Float64,Float64 |    8      8
+test_rrule: two2three on Float64,Float64 |    9      9
```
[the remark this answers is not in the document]
oxinabox: too slow little bot, too slow

PR #182, opened 2021-06-24
unique(TRANSFORMS_TO_ALT_TANGENTS)

oxinabox:
So that if someone `push!` something into this twice, or after we fill this with many things to be tested by default,
but the `push!`'s from before remain,
we don't get duplicates.

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -104,7 +104,7 @@ function test_frule(
     f,
     args...;
     output_tangent=Auto(),
-    tangent_transforms=TRANSFORMS_TO_ALT_TANGENTS,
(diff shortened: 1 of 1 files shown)


mzgubic:
I don't think this would work since it contains functions
```julia
julia> unique([x -> x, x -> x])
2-element Vector{Function}:
 #5 (generic function with 1 method)
 #6 (generic function with 1 method)
```

oxinabox:
true, bother.

Maybe we should call `unique` after applying the transforms then?

mzgubic:
I am not too bothered by this, but we could. 🤷

oxinabox:
closing in favor of #183

PR #186, opened 2021-07-02
WIP: Add the option to test tangent types 

mzgubic:
Companion to https://github.com/JuliaDiff/ChainRules.jl/pull/459.

We might want this? To check that the rule returns the correct tangent types. 

Note that this needs to be made non-breaking by changing the default to `false`. We can flip the sign with the next breaking release.


oxinabox:
I'm not sure.
Possibly for testing this we have something better.
For example checking that you can add it to the primal and the result has structual zeros in the same place the primal does.

PR #189, opened 2021-07-21
Bring back rand_tangent

oxinabox:
Basically reverts #147  bringing it back home,
along with the history it has gained since then


mzgubic (approved):

mzgubic:
ah, also needs a version bump

oxinabox:
I think I might do this as part of v1.0.0?
So that there is no chance of it breaking things unexpectedly from a name-clash?

PR #203, opened 2021-08-05
Use `NoTangent()` instead of `nothing` to designate non-differentiability

mzgubic:
Needed so that https://github.com/JuliaDiff/FiniteDifferences.jl/pull/189 does not break ChainRules tests

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "1.0.0"
+version = "1.1.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


AlexRobson (commented):
This is changing the length of the return - i.e. from (nothing, nothing...) to  a single `NoTangent()`. Is this intended? 

I'm surprised no tests had to change to handle this. 

I think that this means that [here](https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/master/src/testers.jl#L224) given that the output of `fd_cotangents`) will only be length one, the zip will only loop over the first argument.

mzgubic:
well caught, that's a mistake

willtebbutt (approved):
I have no objection to this going in, but like @AlexRobson I'm a bit surprised that there's no changes needed to the tests. Do we just not have great coverage in this package?

mzgubic:
I think the reason is that this is an implementation detail. It is not visible anywhere outside of the package.

PR #205, opened 2021-08-13
don't specialize any functions

simeonschaub:
This takes the time to run the ChainRules test suite from 52 minutes
down to 40 in my tests. I think there are still more gains to be had
here by playing around a bit with `Experimental.@compiler_options`, but
initially that caused some of the inference tests to fail.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "1.1.0"
+version = "1.2.0"
 
 [deps]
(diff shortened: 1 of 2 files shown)


oxinabox:
This makes sense.
So many of the things we do, we do only once.
Since we are testing a particular type combo.

What happenes if we turn it back on for the rand_tangent file?
Those methods we do call a lot with the same types

oxinabox (approved):

simeonschaub:
Good point! Trying that right now.

simeonschaub:
> What happenes if we turn it back on for the rand_tangent file?
Those methods we do call a lot with the same types

Looks like that helps in some tests, but was overall slower. We probably don't spend enough time in `rand_tangent` that this specialization is worth it.

PR #208, opened 2021-08-15
WIP: rrule dot test

willtebbutt:
Please feel free to take charge of this PR if you want to work on this, or to make another one. I can't promise to actively work on it. I just wanted to see what implementing this might look like in practice.

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -172,7 +172,7 @@ function test_rrule(
     config::RuleConfig,
     f,
     args...;
-    output_tangent=Auto(),
+    output_cotangent=Auto(),
     check_thunked_output_tangent=true,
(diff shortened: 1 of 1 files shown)


oxinabox:
Some of the code you deleted we still need.

src/testers.jl
```diff
@@ -172,7 +172,7 @@ function test_rrule(
     config::RuleConfig,
     f,
     args...;
-    output_tangent=Auto(),
+    output_cotangent=Auto(),
```
oxinabox, line 175: So I don't forget if someone does come back to this.
This can't be done.
It is a breaking change.

In general we have decide to use `tangent` as the word to refer to both `tangent`s and `cotangent`s

PR #210, opened 2021-08-20
Improve docs on testing AD gradients

mzgubic:
They were written before `rrule_via_ad` was a thing.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -183,7 +183,8 @@ end
 
 that we do not know an `rrule` for, and we want to check whether the gradients provided by the AD system are correct.
 
-To test gradients the AD system will need to overload `rrule_via_ad` function which wraps the gradients computed by AD.
+To test gradients the AD system you need to provide a `rrule_f` function that acts like calling `rrule` but use AD rather than a defined rule.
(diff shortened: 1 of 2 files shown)


docs/src/index.md
```diff
@@ -183,15 +183,29 @@ end
 
 that we do not know an `rrule` for, and we want to check whether the gradients provided by the AD system are correct.
 
-Firstly, we need to define an `rrule`-like function which wraps the gradients computed by AD.
+To test gradients the AD system will need to overload `rrule_via_ad` function which wraps the gradients computed by AD.
```
oxinabox, line 186: Maybe something like:
```suggestion
To test gradients the AD system you need to provide a `rrule_f` function that acts like calling `rrule` but use AD rather than a defined rule.
This has the exact same semantics as is required to overload `ChainRulesCore.rrule_via_ad`, thus almost all systems doing so should just overload that, and pass in that and the config, and then trigger `test_rrule(MyADConfig, f, xs; rrule_f = ChainRulesCore.rrule_via_ad)`.
```

oxinabox (approved):
This should also link to https://juliadiff.org/ChainRulesCore.jl/stable/config.html

PR #215, opened 2021-09-08
Catch inference problems for thunked cotangents

### LINKED ISSUE #214: Inference tests do not catch problems with thunked inputs
Consider:

```julia
julia> my_id(x) = x
my_id (generic function with 1 method)

julia> function ChainRulesCore.rrule(::typeof(my_id), x)
           my_id_pb(ȳ) = (NoTangent(), ȳ)
           function my_id_pb(ȳ::AbstractThunk)
               precision = rand() > 0.5 ? Float64 : Float32
               return (NoTangent(), precision(unthunk(ȳ)))
           end
           return x, my_id_pb
       end

julia> test_rrule(my_id, 2.0)
Test Summary:                | Pass  Total
test_rrule: my_id on Float64 |    7      7
Test.DefaultTestSet("test_rrule: my_id on Float64", Any[], 7, false, false)
```

while really we should be catching that the thunked cotangents result in type unstable pullbacks. This may seem obscure but is quite relevant because of [closures being unstable if they call another method of themselves](https://juliadiff.org/ChainRulesCore.jl/dev/writing_good_rules.html#Be-careful-about-pullback-closures-calling-other-methods-of-themselves).

mzgubic:
Closes #214 

Note that this might break things since it might catch previously undiscovered bugs. I interpret it as a bug fix however, since we wanted to catch those anyway.

### DIFF
--- a/test/testers.jl
+++ b/test/testers.jl
@@ -723,5 +723,6 @@ end
 
     @testset "inference of thunked cotangents" begin
         @test errors(() -> test_rrule(my_id, 2.0))
+        test_rrule(my_id, 2.0; check_inference=false)
     end
 end
--- a/test/testers.jl
+++ b/test/testers.jl
@@ -57,6 +57,15 @@ end
 abstract type MySpecialTrait end
 struct MySpecialConfig <: RuleConfig{Union{MySpecialTrait}} end
 
+my_id(x) = x
+function ChainRulesCore.rrule(::typeof(my_id), x)
+    my_id_pb(ȳ) = (NoTangent(), ȳ)
+    function my_id_pb(ȳ::AbstractThunk)
+        precision = rand() > 0.5 ? Float64 : Float32
+        return (NoTangent(), precision(unthunk(ȳ)))
+    end
+    return x, my_id_pb
+end
 
 @testset "testers.jl" begin
     @testset "test_scalar" begin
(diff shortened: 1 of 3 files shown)


test/testers.jl
```diff
@@ -57,6 +57,15 @@ end
 abstract type MySpecialTrait end
 struct MySpecialConfig <: RuleConfig{Union{MySpecialTrait}} end
 
+my_id(x) = x
```
oxinabox, line 60: This should be moved inside the `@testset`.
So it is clear what it is for testing.

oxinabox (approved):
If this doesn't break anything that we have reverse dep tests,
or if it only break ChainRules and we can merge the fix first,
I am happy to say this is nonbreaking.

It makes the check_inferred actually check.

test/testers.jl
```diff
@@ -711,4 +720,9 @@ struct MySpecialConfig <: RuleConfig{Union{MySpecialTrait}} end
 
         ChainRulesTestUtils.TEST_INFERRED[] = true
     end
+
+    @testset "inference of thunked cotangents" begin
+        @test errors(() -> test_rrule(my_id, 2.0))
+        test_rrule(my_id, 2.0; check_inferred=false)
```
oxinabox, line 726: test also if that this is fine if we pass `check_thunked_output_tangent=false` ?

test/testers.jl
mzgubic: oh right, it's only the structs that have to be top level in 1.0

PR #216, opened 2021-09-21
make it easier to find out what ⊢is and how to type it

st--:
I'm not sure if this way of trying to add a reference from https://juliadiff.org/ChainRulesTestUtils.jl/dev/index.html to https://juliadiff.org/ChainRulesTestUtils.jl/dev/api.html#ChainRulesTestUtils.:%E2%8A%A2 is correct, but I found it quite frustrating to see a symbol I've not seen before, with no idea how to type it or what it does - at least a cross-link would be useful to save people from having to read through the source code:)

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -134,14 +134,14 @@ test_frule(foo ⊢ Tangent{Foo}(;a=rand()), rand())
 
 ## Specifying Tangents
 [`test_frule`](@ref) and [`test_rrule`](@ref) allow you to specify the tangents used for testing.
-This is done by passing in [`x ⊢ Δx`](@ref :⊢), where `x` is the primal and `Δx` is the tangent, in the place of the primal inputs.
-If this is not done the tangent will be automatically generated via `FiniteDifferences.rand_tangent`.
+By default, tangents will be automatically generated via `FiniteDifferences.rand_tangent`.
+To explicitly specify a tangent, pass in [`x ⊢ Δx`](@ref `⊢`), where `x` is the primal and `Δx` is the tangent, in the place of the primal inputs.
+(You can enter `⊢` via `\vdash` + tab in the Julia REPL and supporting editors.)
 A special case of this is that if you specify it as `x ⊢ NoTangent()` then finite differencing will not be used on that input.
 Similarly, by setting the `output_tangent` keyword argument, you can specify the tangent for the primal output.
 
 This can be useful when the default provided `FiniteDifferences.rand_tangent` doesn't produce the desired tangent for your type.
-For example the default tangent for an `Int` is `NoTangent()`.
-Which is correct e.g. when the `Int` represents a discrete integer like in indexing.
+For example, the default tangent for an `Int` is `NoTangent()`, which is correct e.g. when the `Int` represents a discrete integer like in indexing.
 But if you are testing something where the `Int` is actually a special case of a real number, then you would want to specify the tangent as a `Float64`.
 
 Care must be taken when manually specifying tangents.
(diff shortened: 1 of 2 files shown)


st--:
Side note - would you want the docstrings of e.g. test_frule and test_rrule to also more explicitly tell the users what ⊢ does and how to type it other than select-copy-paste ?

willtebbutt:
This is a helpful addition.

> Side note - would you want the docstrings of e.g. test_frule and test_rrule to also more explicitly tell the users what ⊢ does and how to type it other than select-copy-paste ?

Yes please -- I find having whatever I'm meant to type in `\command` format really helpful (I'm not actually sure what that is in this situation). @mzgubic @st-- do you recall the command is in this situation?

devmotion:
IIRC it is `\vdash`.

mzgubic:
I often forget this too. Only recently I've learned that it is possible to find it out in the help mode:
```julia
help?> ⊢
"⊢" can be typed by \vdash<tab>

search:

  No documentation found.

  Binding ⊢ does not exist.
```

mzgubic (approved):

docs/src/index.md
```diff
@@ -134,7 +134,7 @@ test_frule(foo ⊢ Tangent{Foo}(;a=rand()), rand())
 
 ## Specifying Tangents
 [`test_frule`](@ref) and [`test_rrule`](@ref) allow you to specify the tangents used for testing.
-This is done by passing in `x ⊢ Δx`, where `x` is the primal and `Δx` is the tangent, in the place of the primal inputs.
+This is done by passing in [`x ⊢ Δx`](@ref :⊢), where `x` is the primal and `Δx` is the tangent, in the place of the primal inputs.
```
st--, line 137:
```suggestion
This is done by passing in [`x ⊢ Δx`](@ref :⊢), where `x` is the primal and `Δx` is the tangent, in the place of the primal inputs.
You can type `⊢` in the Julia REPL using `\vdash<tab>`
```

src/testers.jl
```diff
@@ -144,21 +147,22 @@ end
 
 # Arguments
 - `config`: defaults to `ChainRulesTestUtils.ADviaRuleConfig`.
-- `f`: Function to which rule should be applied. Can also provide `f ⊢ f̄`.
-- `args` either the primal args `x`, or primals and their tangents: `x ⊢ x̄`
-    - `x`: input at which to evaluate `f` (should generally be set to an arbitary point in the domain).
-    - `x̄`: currently accumulated cotangent, will be generated automatically if not provided
+- `f`: function for which the `rrule` should be tested. Its tangent can be provided using `f ⊢ f̄`.
```
st--, line 150: Should this also be _co_tangent?
```suggestion
- `f`: function for which the `rrule` should be tested. Its cotangent can be provided using `f ⊢ f̄`.
```

src/testers.jl
```diff
@@ -144,21 +147,22 @@ end
...
+  (You can enter `⊢` via `\\vdash` + tab in the Julia REPL and supporting editors.)
+- `args...`: either the primal args `x`, or primals and their tangents: `x ⊢ x̄`
```
st--, line 152:
```suggestion
- `args...`: either the primal args `x`, or primals and their cotangents: `x ⊢ x̄`
```
? (for consistency with https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/216/files#diff-988b853ad3c22aebac9421d3a31e60f9330789df43bccfb9fb75365048398fbaR154)

st--:
@mzgubic @willtebbutt other than confirming whether it should be contangents or not, is there anything else missing before this can get merged?

mzgubic:
Technically they are cotangents but we often do not make a distinction. E.g. `Tangent` type is used for both tangents and cotangents. Either is fine with me in the docstring.

I would be in favour of a patch bump, but don't feel strongly about it. I'll just merge it to avoid running CI again

Issue #217, opened 2021-09-21
How to test rules with @not_implemented arguments?


st--:
If I write a, say, `rrule` that returns NoTangent() for some of the arguments, then `test_rrule` ignores those arguments and everything is fine. However, if I want to be more explicit and make use of `@not_implemented("...")` for some of the arguments, then `test_rrule` fails with
```
test_rrule: newton_inner_loop on Matrix{Float64},typeof(dist_y_given_f),Vector{Bool}: Test Failed at ~/.julia/packages/ChainRulesTestUtils/KlXK9/src/testers.jl:233
  Expression: ad_cotangent isa NoTangent
   Evaluated: NotImplemented(<...>) isa NoTangent
```

How am I supposed to test rules for which I want to make use of `@not_implemented`?
Should https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/a46fbbc1348c4ab1dc455cf5352f972295de967e/src/testers.jl#L233  get wrapped in an if clause that makes the line also pass if ad_cotangent isa NotImplemented ? (Happy to submit a PR including a regression test if that's the way to go.)

oxinabox:
What was the call to `test_rrule` you made?
Did you pass a thing telling `test_rrule` something was `NoTangent` when actually it was `NotImplemented`? Those are distinct things and so it would be correct to fail.

I think it should just work: there is logic in `test_approx` that checks for `NotImplemented`
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/master/src/check_result.jl

But there could well be bugs in it.

A MWE would be useful.

devmotion:
It should just work: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/a46fbbc1348c4ab1dc455cf5352f972295de967e/test/testers.jl#L642-L647

But maybe I missed something, so a MWE would be good :slightly_smiling_face:

devmotion:
Another example are the tests in SpecialFunctions (the definitions there use `@not_implemented` since some derivatives can't be implemented without a lot of effort and/or introducing additional dependencies) which don't have to handle `@not_implemented` in a special way: https://github.com/JuliaMath/SpecialFunctions.jl/blob/master/test/chainrules.jl

st--:
MRE:
```julia
using ChainRulesCore
using ChainRulesTestUtils

function foo(K, fun)
    return 2K
end

[...]

PR #218, opened 2021-09-21
Allow NotImplemented tangents for things that have a correct tangent of NoTangent

devmotion:
This is one possible approach to fix https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/217.

An alternative would be to tell users to define `rand_tangent` or, probably better, pass a suitable tangent of type `NotImplemented`. It is a bit inconvenient though since currently we check for equality, including the message AND the LineNumberNodes. I.e., one can't just use `test_rrule(f, x \vdash @not_implemented("does not work"))` since the LineNumberNode would be different from the one used inside the `rrule`. Maybe this should be changed and we should only check if the messages are equal?

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -317,7 +317,7 @@ function _test_cotangent(
     # this situation can occur if a cotangent is not implemented and
     # the default `rand_tangent` is `NoTangent`: e.g. due to having no fields
     # the `@test_broken` below should tell them that there is an easy implementation for
-    # for this case of `NoTangent()` (`@test_broken false` would be less useful!)
+    # this case of `NoTangent()` (`@test_broken false` would be less useful!)
     # https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/217
     @test_broken ad_cotangent isa NoTangent
 end
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -220,33 +220,8 @@ function test_rrule(
         # Correctness testing via finite differencing.
         is_ignored = isa.(accum_cotangents, NoTangent)
         fd_cotangents = _make_j′vp_call(fdm, call, ȳ, primals, is_ignored)
-
-        for (accum_cotangent, ad_cotangent, fd_cotangent) in zip(
-            accum_cotangents, ad_cotangents, fd_cotangents
-        )
-            if accum_cotangent isa NoTangent  # then we marked this argument as not differentiable
-                @assert fd_cotangent === NoTangent()
-                ad_cotangent isa ZeroTangent && error(
-                    "The pullback in the rrule should use NoTangent()" *
-                    " rather than ZeroTangent() for non-perturbable arguments.",
-                )
-                if ad_cotangent isa ChainRulesCore.NotImplemented
-                    # this situation can occur if a cotangent is not implemented and
-                    # the default `rand_tangent` is `NoTangent`: e.g. due to having no fields
-                    # the `@test_broken` bellow should tell them that their is an easy
-                    # implementation for this case of `NoTangent()`
-                    # https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/217
-                    @test_broken ad_cotangent isa NoTangent
-                else
-                    @test ad_cotangent isa NoTangent  # we said it wasn't differentiable.
-                end
-            else
-                ad_cotangent isa AbstractThunk && check_inferred && _test_inferred(unthunk, ad_cotangent)
-
-                # The main test of the actual derivative being correct:
-                test_approx(ad_cotangent, fd_cotangent; isapprox_kwargs...)
-                _test_add!!_behaviour(accum_cotangent, ad_cotangent; isapprox_kwargs...)
-            end
+        foreach(accum_cotangents, ad_cotangents, fd_cotangents) do args...
+            _test_cotangent(args...; check_inferred=check_inferred, isapprox_kwargs...)
         end
 
         if check_thunked_output_tangent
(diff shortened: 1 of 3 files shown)


src/testers.jl
```diff
@@ -230,7 +230,14 @@ function test_rrule(
                     "The pullback in the rrule should use NoTangent()" *
                     " rather than ZeroTangent() for non-perturbable arguments.",
                 )
-                @test ad_cotangent isa NoTangent  # we said it wasn't differentiable.
+                if ad_cotangent isa ChainRulesCore.NotImplemented
+                    # this situation can occur if a cotangent is not implemented and
+                    # the default `rand_tangent` is `NoTangent`:
```
oxinabox, line 235:
```suggestion
                    # the default `rand_tangent` is `NoTangent`: e.g. due to having no fields
                    # the `@test_broken` bellow should tell them that their is an easy
                    # implementation for this case of `NoTangent()`
```

oxinabox (approved):
I am convinced that this is the best way.

Do you want to extract some of this code into a helper method?
Its getting long.
Maybe a function for the content of the branch:  `accum_cotangent isa NoTangent` ?
Or maybe a method that dispatches on `accum_cotangent`, and `ad_cotangent`  ?

Regardless, change as you will then merge and tag when happy.

src/testers.jl
```diff
@@ -230,7 +230,16 @@ function test_rrule(
                     "The pullback in the rrule should use NoTangent()" *
                     " rather than ZeroTangent() for non-perturbable arguments.",
                 )
-                @test ad_cotangent isa NoTangent  # we said it wasn't differentiable.
+                if ad_cotangent isa ChainRulesCore.NotImplemented
+                    # this situation can occur if a cotangent is not implemented and
+                    # the default `rand_tangent` is `NoTangent`: e.g. due to having no fields
+                    # the `@test_broken` bellow should tell them that their is an easy
```
st--, line 236:
```suggestion
                    # the `@test_broken` below should tell them that there is an easy
```

st--: what do you mean by "there is an easy implementation"?

oxinabox: The correct implementation of anything that `test_rrule` determined was a `NoTangent` is almost certainly `NoTangent()`
Which is easy.
It might not generalize to cases that are not tested, but for the types passed into this test it is (almost certainly) correct.

devmotion: I fixed the spelling mistakes in the latest commit (I had already some major changes in my local branch so I just fixed it there).

src/testers.jl
```diff
@@ -285,3 +269,58 @@ function _is_inferrable(f, args...; kwargs...)
...
+    ::ChainRulesCore.NotImplemented,
+    ::NoTangent;
+    kwargs...,
+)
+    # this situation can occur if a cotangent is not implemented and
+    # the default `rand_tangent` is `NoTangent`: e.g. due to having no fields
+    # the `@test_broken` below should tell them that there is an easy
+    # implementation for this case of `NoTangent()`
+    # https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/217
+    @test_broken false
```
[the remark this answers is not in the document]
devmotion: Ah true, I thought I could simplify the code but I see that it was not a good idea :smile:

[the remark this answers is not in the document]
oxinabox: Macro's they will do that to you

src/testers.jl
```diff
@@ -285,3 +269,58 @@ function _is_inferrable(f, args...; kwargs...)
...
+)
+    # this situation can occur if a cotangent is not implemented and
+    # the default `rand_tangent` is `NoTangent`: e.g. due to having no fields
+    # the `@test_broken` below should tell them that there is an easy implementation for
+    # this case of `NoTangent()` (`@test_broken false` would be less useful!)
+    # https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/217
+    @test_broken ad_cotangent isa NoTangent
+end
+function _test_cotangent(::NoTangent, ad_cotangent, fd_cotangent; kwargs...)
+    error("cotangent obtained with finite differencing has to be NoTangent()")
```
st--, line 325: In what circumstances might this be hit? As a user of ChainRulesTestUtils I would find this error message quite confusing, if you give me a bit more context maybe I can suggest a rephrasing:)

src/testers.jl
```diff
@@ -285,3 +269,58 @@ function _is_inferrable(f, args...; kwargs...)
...
+    kwargs...,
+)
+    ad_cotangent isa AbstractThunk && check_inferred && _test_inferred(unthunk, ad_cotangent)
+
+    # The main test of the actual derivative being correct:
+    test_approx(ad_cotangent, fd_cotangent; kwargs...)
+    _test_add!!_behaviour(accum_cotangent, ad_cotangent; kwargs...)
+end
+
+# we marked the argument as non-differentiable
```
st--, line 301:
```suggestion
# if accum_cotangent isa NoTangent, then we marked the argument as non-differentiable
```

src/testers.jl
```diff
@@ -285,3 +269,58 @@ function _is_inferrable(f, args...; kwargs...)
...
+function _test_cotangent(::NoTangent, ad_cotangent, ::NoTangent; kwargs...)
+    @test ad_cotangent isa NoTangent
+end
+function _test_cotangent(::NoTangent, ::ZeroTangent, ::NoTangent; kwargs...)
+    error(
+        "The pullback in the rrule should use NoTangent()" *
+        "rather than ZeroTangent() for non-perturbable arguments."
```
st--, line 308:
```suggestion
        " rather than ZeroTangent() for non-perturbable arguments."
```

src/testers.jl
```diff
@@ -285,3 +269,58 @@ function _is_inferrable(f, args...; kwargs...)
...
+    ::NoTangent,
+    ad_cotangent::ChainRulesCore.NotImplemented,
+    ::NoTangent;
+    kwargs...,
+)
+    # this situation can occur if a cotangent is not implemented and
+    # the default `rand_tangent` is `NoTangent`: e.g. due to having no fields
+    # the `@test_broken` below should tell them that there is an easy implementation for
+    # this case of `NoTangent()` (`@test_broken false` would be less useful!)
+    # https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/217
```
st--, line 321:
```suggestion
    # This situation can occur if a cotangent is marked as `@not_implemented` and
    # the default `rand_tangent` is `NoTangent`: e.g. due to having no fields.
    # The `@test_broken` below should tell them that there is an easy implementation for
    # this case of `NoTangent()` (`@test_broken false` would be less useful!)
    # https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/217
```

src/testers.jl
```diff
@@ -285,3 +269,58 @@ function _is_inferrable(f, args...; kwargs...)
...
+    @test_broken ad_cotangent isa NoTangent
```
st--, line 322: > the `@test_broken` below should tell them that there is an easy implementation for this case of `NoTangent()`

I don't understand what this is supposed to say, could you explain so we can make it a more helpful comment for future readers of the codebase?

st--: From the comment by @oxinabox on the outdated diff:

> The correct implementation of anything that test_rrule determined was a NoTangent is almost certainly NoTangent()
Which is easy.
It might not generalize to cases that are not tested, but for the types passed into this test it is (almost certainly) correct.

But I'm still not clear what I'd be expected to do.

src/testers.jl
devmotion: Users will never see it, it's just a dispatch version of the `@assert` statement. It can't happen since we compute `fd_cotangent`s in `test_rrule` and explicitly exclude the arguments with a `NoTangent()` tangent.

st--: E.g., does this suggest a bug in the finite differencing (upstream) code, or a bug in my downstream code?

src/testers.jl
devmotion: If the argument is actually differentiable and `NoTangent` is just caused by an incorrect `rand_tangent` default, then you should provide a correct tangent.

src/testers.jl
devmotion: You will never see it. There's just no way you can end up here: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/f57fc7e827658bd4a00e79cb3c75d99ba292b490/src/testers.jl#L221-L222

So I really don't think we have to worry about downstream users, it is just a comment and a bit more explicit and easier to understand then the current `@assert` statement.

src/testers.jl
devmotion: This is explained in the docstring above.

src/testers.jl
st--: Where would we have ended up if `fun` _had_ fields? How would I tell the dispatcher that an rrule would only apply to functions with no field?

st--: If my example
```julia
function ChainRulesCore.rrule(::typeof(foo), K, fun)
    f = foo(K, fun)

    function pullback(Δf)
        ∂self = NoTangent()
        ∂K = @thunk(2Δf)
        ∂fun = NoTangent()
        return (∂self, ∂K, ∂fun)
    end

    return f, pullback
end
```
should be generic for `fun` (the derivative for fields of `fun` _would_ exist, I just haven't worked it out yet), is `@not_implemented` not the right thing to do ?

devmotion: In general, it is not the right thing to do, the right thing is to implement the derivatives :stuck_out_tongue:  `@not_implemented` is a workaround until you've done this, and therefore all tests with `NotImplemented` use `@test_broken` to indicate that something is broken and should be fixed.

PR #220, opened 2021-10-01
Accept that NaN ≈ NaN in test_approx

oxinabox:
I think this is the behavior we want.
If you get a actual value of NaN and you were expecting a NaN then all seems well.
If both the AD rule and the finite differencing take you here all is well.
(not ok if only one does)

### DIFF
--- a/src/check_result.jl
+++ b/src/check_result.jl
@@ -21,7 +21,7 @@ function test_approx(
     msg="";
     kwargs...,
 )
-    @test_msg msg isapprox(actual, expected; kwargs...)
+    @test_msg msg isapprox(actual, expected; nans=true, kwargs...)
 end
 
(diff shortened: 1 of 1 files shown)


oxinabox:
This is not quite the same as https://github.com/JuliaDiff/FiniteDifferences.jl/issues/192 
which woulderror if the **input** is not finite.
Here we are allow **output** to be NaN,
so we could do both

mzgubic (approved):
needs a version bump and some tests?

oxinabox:
idk if this is a good idea or not

mzgubic:
The only thing I can think of going wrong is getting a `NaN` for the wrong reason. But seems relatively unlikely (and harmless in the end, at least at the tested point?)

PR #224, opened 2021-11-09
Fix tangent type of arrays of (named) tuples from FD

devmotion:
Currently, arrays of tuples and named tuples as a result from FiniteDifferences are not converted to the correct `Tangent` types. This causes the issue in https://github.com/JuliaMath/ChangesOfVariables.jl/pull/4#discussion_r745556709 and is fixed by this PR (confirmed locally).

~~Deeply nested arrays of (named) tuples will still have an incorrect, so maybe a (better?) alternative would be to always recurse arrays until one hits a `AbstractArray{<:Real}` (which would be returned unmodified, also to avoid copies in this case) or a non-array base case such as `Tuple`, `NamedTuple`, and `Any`.~~

The PR replaces the `_maybe_fix_to_composite` with `ProjectTo`.

It also seems that the function is not tested currently. Should I add some tests?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "1.2.3"
+version = "1.2.4"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -16,21 +16,21 @@ uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
 [[ChainRulesCore]]
 deps = ["Compat", "LinearAlgebra", "SparseArrays"]
-git-tree-sha1 = "bdc0937269321858ab2a4f288486cb258b9a0af7"
+git-tree-sha1 = "f885e7e7c124f8c92650d61b9477b9ac2ee607dd"
 uuid = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
-version = "1.3.0"
+version = "1.11.1"
 
 [[ChainRulesTestUtils]]
 deps = ["ChainRulesCore", "Compat", "FiniteDifferences", "LinearAlgebra", "Random", "Test"]
 path = ".."
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "1.2.1"
+version = "1.2.4"
 
 [[Compat]]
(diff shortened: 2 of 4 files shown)


oxinabox:
I wonder if we can do this with `ProjectTo` ?

devmotion:
> I wonder if we can do this with `ProjectTo` ?

Nice, this seems to fix the issue in ChangesOfVariables as well. Great if we can use existing functionality. I'm currently checking if it breaks any tests in CRTestUtils (and if successful, I'll push and we can see if it breaks any downstream packages).

devmotion:
Hmm, tests passed locally but it seems doctests fail since `ProjectTo(::Tuple{Float64,Float64,Float64})` is not defined :cry:

oxinabox:
> @oxinabox would it be OK to add a `ProjectTo(::Tuple)` definition in ChainRulesCore? Or was it omitted on purpose?

it is here, no?
https://github.com/JuliaDiff/ChainRulesCore.jl/blob/99d56b145bb4829931c542e720a015d938efeee4/src/projection.jl#L291

devmotion:
Oh wow, now I'm confused. The error message clearly stated that it is missing: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/runs/4153868239?check_suite_focus=true#step:4:214

devmotion:
Ah the docs use an old version of CRCore: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/runs/4153868239?check_suite_focus=true#step:4:57

oxinabox:
Looks good to me.
Assuming integration tests pass.
Once you **add a test** you can merge and tag when happy.

oxinabox (approved):

devmotion:
There's one test error in ChainRules: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/runs/4154680040?check_suite_focus=true#step:6:162 It is caused by the test in https://github.com/JuliaDiff/ChainRules.jl/blob/edf3a1f48fb5c9af01820aeca6ced94d4f97fa1a/test/rulesets/Base/array.jl#L35

More concretely, the problem is that
```julia
julia> using ChainRulesCore # v 1.11.1

julia> ProjectTo((; y = randn(3)))
identity (generic function with 1 method)
```
I guess this is a bug and has to be fixed in ChainRulesCore since it differs e.g. from
```julia
julia> ProjectTo((randn(3),))
ProjectTo{Tangent{Tuple{Vector{Float64}}, T} where T}(elements = (ProjectTo{AbstractArray}(element = ProjectTo{Float64}(), axes = (Base.OneTo(3),)),),)
```
which projects to a `Tangent`.

devmotion:
I assume the ChainRules.jl test errors will be fixed by https://github.com/JuliaDiff/ChainRules.jl/pull/550.

oxinabox:
yep, merge when happy

PR #226, opened 2021-11-16
Simply rand_tangent(::AbstractArray)

oxinabox:
1. Since https://github.com/JuliaDiff/ChainRulesCore.jl/pull/424/files  `ProjectTo` handles the collapsing of `AbstractArray{NoTangent}` into `NoTangent()`.
2. Further, `ProjectTo(::Array{Float64,0})(::Float64)` handles converting a scalar into a 0 dimensional array.
3. Finally, there is no need for a `collect` since we can make the broadcasting do that (and that might have been specialized by the container and that is fine)

So we can delete all special cases for generating arrays.

This is a minor bump since we are dropping compat with CRC 1.0 and 1.1
I am pretty sure this has no functional change. Though there might be some edge cases.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "1.2.3"
+version = "1.3.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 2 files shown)


mzgubic:
I think the difference is that in this package compression only works on `NoTangent()`, but `ProjectTo` works on `AbstractZero`. So I thought there was a difference. Turns out that it's not the case
```julia
julia> rand_tangent([ZeroTangent(), ZeroTangent()])
NoTangent()
```
because 
```julia
julia> rand_tangent(ZeroTangent())
NoTangent()
```
so there indeed does not seem to be a functional difference.

mzgubic (approved):

PR #228, opened 2022-01-14
Add method for checking rules added to DataType etc

oxinabox:
This puts https://github.com/JuliaDiff/ChainRules.jl/pull/563 into a function. So people can use it in their own modules

Copying my openning post from there

> 
> This should let us catch errors like
> https://github.com/JuliaDiff/ChainRules.jl/pull/562
> before they are merged.
> 
> idk if this is the best way, feedback appreciated.
> 
> 
> The output this will show is:
> ```
> ┌ Error: Bad constructor rrule. typeof(T)` not `Type{T}`
> │   method = rrule(::UnionAll, x::AbstractArray, perm) in ChainRules at /home/oxinabox/JuliaEnvs/ChainRulesWorld/ChainRules.jl/src/rulesets/Base/array.jl:84
> └ @ Main ~/JuliaEnvs/ChainRulesWorld/ChainRules.jl/test/global_state.jl:24
> Test Failed at /home/oxinabox/JuliaEnvs/ChainRulesWorld/ChainRules.jl/test/global_state.jl:25
>   Expression: false
> ERROR: LoadError: There was an error during testing
> in expression starting at /home/oxinabox/JuliaEnvs/ChainRulesWorld/ChainRules.jl/test/global_state.jl:14
> ```

TODO:

- [x] Add to docs
- [x] Test it (will need to add and delete from the method table)


docs/src/index.md
```diff
@@ -9,7 +9,13 @@ For information about ChainRules, including how to write rules, refer to the gen
 [![](https://img.shields.io/badge/docs-main-blue.svg)](https://JuliaDiff.github.io/ChainRulesCore.jl/dev)
 [![](https://img.shields.io/badge/docs-stable-blue.svg)](https://JuliaDiff.github.io/ChainRulesCore.jl/stable)
 
-## Canonical example
+## Testing Method Table Sensibility
+A basic feature of ChainRulesTestUtils is it's ability to check that the method table for `rrule` and `frule` remains sensible.
+This searches the method table for methods that should not exist and when it fails tells you where they were defined.
```
nickrobinson251, line 14: let's be consistent here and throughout on if we're talking about "the method table" (singular) or "the method tables (for `frule` and `rrule`)" (plural). I'm not sure which is more accurate but i think consistency helps to just even follow what's being said :)

since  the function itself is "`test_method_tables_sensibility`" i guess we're going with plural:
```suggestion
A basic feature of ChainRulesTestUtils is its ability to check that the method tables for `rrule` and `frule` remain sensible.
This searches the method tables for methods that should not exist and when it fails tells you where they were defined.
```

src/global_checks.jl
```diff
@@ -0,0 +1,56 @@
...
+_parameters(sig::DataType) = sig.parameters
+_parameters(sig::Union) = Base.uniontypes(sig)
+
+"""
+    test_method_tables_sensibility()
+
+Checks that the method tables for `rrule` and `frule` are sensible.
+This in future may carry out a number of checks, but presently just checks to make sure that
+no rules have been added to the very general `DataType`, `Union` or `UnionAll` types.
+This is easy to do when writing rules for constructors.
```
nickrobinson251, line 16:
```suggestion
no rules have been added to the very general `DataType`, `Union` or `UnionAll` types,
which is easy to do accidentally when writing rules for constructors.
```

src/global_checks.jl
```diff
@@ -0,0 +1,56 @@
...
+It happens if you writeg. `rrule(::typeof(Foo), x)` rather than `rrule(::Type{<:Foo}, x)`:
```
nickrobinson251, line 17:
```suggestion
It happens if you write e.g. `rrule(::typeof(Foo), x)` rather than `rrule(::Type{<:Foo}, x)`.
```

src/global_checks.jl
```diff
@@ -0,0 +1,56 @@
...
+    _parameters(type)
+Extracts the type-parameters of the `type`.
+e.g. `_parameters(Foo{A, B, C}) == [A, B, C]`
+"""
+_parameters(sig::UnionAll) = _parameters(sig.body)
+_parameters(sig::DataType) = sig.parameters
+_parameters(sig::Union) = Base.uniontypes(sig)
+
+"""
+    test_method_tables_sensibility()
```
nickrobinson251, line 11: is "sensibility" really the word, rather than just "sense"?

Also maybe just
```suggestion
    test_method_tables()
```
or even have separate methods
```suggestions
    test_method_table(frule|rrule)
```

src/global_checks.jl
```diff
@@ -0,0 +1,56 @@
...
+            function_type = if method.sig <: Tuple{Any, RuleConfig, Type, Vararg}
+                _parameters(method.sig)[3]
+            elseif method.sig <: Tuple{Any, Type, Vararg}
+                _parameters(method.sig)[2]
+            else
+                nothing
+            end
+            
+            @test_msg(
+                "Bad constructor rrule. typeof(T) used rather than `Type{T}`. $method",
```
nickrobinson251, line 36:
```suggestion
                "Bad constructor rrule. `typeof(T)` used rather than `Type{T}`. $method",
```

src/global_checks.jl
```diff
@@ -0,0 +1,56 @@
...
+            function_type = if method.sig <: Tuple{Any, RuleConfig, Any, Type, Vararg}
+                _parameters(method.sig)[4]
+            elseif method.sig <: Tuple{Any, Any, Type, Vararg}
+                @show _parameters(method.sig)[3]
+            else
+                nothing
+            end
+
+            @test_msg(
+                "Bad constructor frule. typeof(T) used rather than `Type{T}`. $method",
```
nickrobinson251, line 51:
```suggestion
                "Bad constructor frule. `typeof(T)` used rather than `Type{T}`. $method",
```

src/global_checks.jl
```diff
@@ -0,0 +1,56 @@
...
+
+Checks that the method tables for `rrule` and `frule` are sensible.
+This in future may carry out a number of checks, but presently just checks to make sure that
+no rules have been added to the very general `DataType`, `Union` or `UnionAll` types.
+This is easy to do when writing rules for constructors.
+It happens if you writeg. `rrule(::typeof(Foo), x)` rather than `rrule(::Type{<:Foo}, x)`:
+This would then actually define `rrule(::DataType, x)`. (or `UnionAll` if `Foo`
+was parametric, or `Union` if `Foo` was a type alias for a `Union`)
+"""
+function test_method_tables_sensibility()
```
oxinabox, line 21:
```suggestion
function test_method_tables()
```

mzgubic (commented):
I think the logic is fine and LGTM for implementation. However, I still worry about people writing rules for their package, testing the rules with CRTU, and thinking they are fine because they didn't run the CR tests.

Could we refactor what we currently have in a function `test_method_sig(rule_type, sig)`, and then call it in `test_method_tables`. In addition, we can then call `test_method_sig` inside the `test_rrule`, where we construct the `sig` manually?

oxinabox:
> I think the logic is fine and LGTM for implementation. However, I still worry about people writing rules for their package, testing the rules with CRTU, and thinking they are fine because they didn't run the CR tests.

They don't need to run CR's tests they just need to run the `test_method_tables()` in their tests.
which we do document in this PR as something everyone should do.

> Could we refactor what we currently have in a function `test_method_sig(rule_type, sig)`, and then call it in `test_method_tables`. In addition, we can then call `test_method_sig` inside the `test_rrule`, where we construct the `sig` manually?

I don't actually think we can construct the `sig` manually.
Because julia functions are called with concrete types and `typeof(Foo)` is both `DataType` and `Type{Foo}`.
We need to look at the method tables.
I guess with could call `which(rrule(f, args...)` to do that in each rule though.
Do you think that is better?

mzgubic:
>They don't need to run CR's tests they just need to run the test_method_tables() in their tests.
which we do document in this PR as something everyone should do.

Yeah, I agree it _should_ be clear that doing `::typeof(Foo)` is bad from the docs, and yet it happens sometimes. I wouldn't rely on people reading the docs to add this new function to their tests 🤷.

>I guess with could call which(rrule, (f, args...)) to do that in each rule though.
Do you think that is better?

I would be strongly in favour of adding this check to the `test_rrule`. We should be making life easy for people who are using ChainRulesCore, and we can do this here by:
- not needing them to read the testing section of the docs
- automatically throw an informative error message if they mess up and have not read the testing docs.

mcabbott:
If you run it in `test_rrule`, you can say which rule exactly is the problem, too. (Although still not the line number...)

oxinabox:
> If you run it in test_rrule, you can say which rule exactly is the problem, too. (Although still not the line number...)

We actually say the line number in the error message it is printed as part of printing the2 `Method` object.
I wonder if there are other places we can use that,

mcabbott:
Oh I guess you can get the method's line number, which is more useful than the tests's, for this purpose.

oxinabox:
oh it is surprisingly hard to do this inside the `test_frule` / `test_rrule` functions.
Because we don't know if the config will be used or not (if not it will redispatch).
also we might not be using `frule` or `rrule` in the first place because they might be passing in the `rrule_f` and `frule_f` functions to not test a `frule` or `rrule` at all.

I am tempted to just put in the global check, and get this in, and then later we can add the per thing check if we find people often don't realise there is a global check

oxinabox:
added back global check

mzgubic:
ok, let's 🚀

mzgubic (approved):

PR #230, opened 2022-01-17
Fix test_approx() when one of the arguments is a broadcasted multidimensional array

dfdx:
`collect()` on broadcasted arrays doesn't preserve its shape, but instead creates a flat array. This breaks tests e.g. for most activation functions in NNlib:

```julia
test_rrule(Broadcast.broadcasted, NNlib.σ, rand(3, 4))
```
which produces:
```
Test threw exception
  DimensionMismatch("dimensions must match: a has dims (Base.OneTo(2), Base.OneTo(2)), b has dims (Base.OneTo(4),), mismatch at 1")
```


oxinabox:
Should this be fixed up stream in JuliaLang/Julia
https://github.com/JuliaLang/julia/issues/43847 ?

dfdx:
Amazing analysis! Let's see how it turns out in the `Base` and whether the fix gets to all popular Julia versions. If it's not applied to any of the versions supported in ChainRules, we can still apply this PR to be on the safe side.

oxinabox:
Cool so this has now been addressed in julia v1.9.0
Unclear as to if this will be backported,
but we can add a comment to the source saying: 
`#Works around https://github.com/JuliaLang/julia/issues/43847 on pre-Julia v1.9`
to this PR and then merge this if we still think this is needed?

dfdx:
Yeah, since we support older versions of Julia, I believe it's still worth to have this workaround.
I added the suggested comment to the PR.

oxinabox (approved):

PR #236, opened 2022-01-21
NoTangent to Ptr

oxinabox:
Should fix cases where automatically generating tangents is failing because object contains a Ptr


mzgubic (approved):
bump the version but LGTM

Issue #234, opened 2022-01-17
More errors with broadcasted array


dfdx:
```julia
test_rrule(Broadcast.broadcasted, NNlib.relu, rand(3,4); check_inferred=false)
```
output:
```
test_rrule: broadcasted on typeof(relu),Matrix{Float64}: Test Failed at /home/andrey/.julia/packages/ChainRulesTestUtils/NGSQO/src/check_result.jl:24
  Expression: isapprox(actual, expected; kwargs...)
  Problem:  Matrix{ChainRulesCore.AbstractTangent}[1]
   Evaluated: isapprox(0.0, 0.2999999999999697; rtol = 1.0e-9, atol = 1.0e-9)
...
```
I see that `ad_cotangents` is pretty weird:

```julia
julia> unthunk(ad_cotangents[3])
3×4 Matrix{ChainRulesCore.AbstractTangent}:
 NoTangent()                                                                                           …  NoTangent()
 Tangent{Tuple{Matrix{Float64}}}([0.97 -0.79 -2.63 2.3; 3.82 -7.89 8.3 2.61; 8.51 -2.73 -7.34 5.88],)     Tangent{Tuple{Matrix{Float64}}}([0.97 -0.79 -2.63 2.3; 3.82 -7.89 8.3 2.61; 8.51 -2.73 -7.34 5.88],)
 NoTangent()                                                                                              NoTangent()
```
I _think_ this is because `rand_tangent()` treats `Broadcasted` as structs and creates `Tangent` instead of creating an array or another `Broadcasted`:

```julia
julia> ȳ = rand_tangent(broadcasted(NNlib.relu, rand(3,4)))
Tangent{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(relu), Tuple{Matrix{Float64}}}}(f = NoTangent(), args = Tangent{Tuple{Matrix{Float64}}}([-8.6 0.92 2.42 -4.99; 6.19 -6.98 2.23 -2.35; -0.66 4.4 -5.54 2.68],), axes = NoTangent())
```
But this is getting a bit too deep for me to understand if my hypothesis is correct and find an appropriate solution :grimacing:

mcabbott:
One trick would be to define a rule for `collect∘Broadcast.broadcasted` or something, and test that. Mathematically the same, but an easier object for CRTU to handle?

[...]

PR #238, opened 2022-01-24
Add rand_tangent() for Broadcasted

dfdx:
Motivation is in #234. Note that this PR doesn't solve the described issue completely due to #230, but even by itself should be useful.


oxinabox (approved):

oxinabox:
thanks! makes sense to me

Issue #241, opened 2022-02-03
Add a config for testing rules that call back into AD without an frule/rrule


sethaxen:
Suppose have a function `foo` and rules:
```julia
foo(f, x) = f(x)
function rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(foo), f, x)
    y, back = rrule_via_ad(config, f, x)
    foo_pullback(dx) = (NoTangent(), back(dx)...)
    return y, foo_pullback
end
function frule(config::RuleConfig{>:HasForwardMode}, (_, df, dx), ::typeof(foo), f, x)
    y, dy = frule_via_ad(config, (df, dx), f, x)
    return y, dy
end
```

Now we want to test these rules. But suppose no `frule` or `rrule` exists currently for `f`. Then we need to load an AD. But suppose, as is the case for `frule`, no working AD currently uses `frule`s. Then we have to write our own `frule` for `f` using the AD/FD of our choice.

It would be convenient if we instead provided a FiniteDifferences config for which `frule_via_ad` and `rrule_via_ad` used FD. Then, the user could provide this config to test such rules.

Potentially this could be conceived as some sort of reworking of  `ADviaRuleConfig`.

sethaxen:
No I'm not working on it.

Issue #221, opened 2021-10-01
Add finite differencing based RuleConfigs


oxinabox:
Right now for testing we use the `ADviaRuleConfig` which only works if we actually have a `rrule`/`frule` for the part.
Which is generally good for testing as it avoids adding any extra noise
but sometimes it is too hard to write the rule for the thing being tested.
It would be good to also support doing so via finite differencing.

PR #243, opened 2022-02-15
Implement ADviaFDConfig

mzgubic:
Closes https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/241 and https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/221

### DIFF
--- a/src/rule_config.jl
+++ b/src/rule_config.jl
@@ -38,10 +38,9 @@ function ChainRulesCore.rrule_via_ad(config::ADviaFDConfig, f, args...; kws...)
     call(f, xs...) = f(xs...; kws...)
 
     # this block is here just to work out which tangents should be ignored
-    primals_and_tangents = auto_primal_and_tangent.((f, args...))
-    primals = primal.(primals_and_tangents)
-    accum_cotangents = tangent.(primals_and_tangents)
-    is_ignored = isa.(accum_cotangents, NoTangent)
+    primals = (f, args...)
+    primals_and_tangents = auto_primal_and_tangent.(primals)
+    is_ignored = isa.(tangent.(primals_and_tangents), NoTangent)
 
     function f_pb(ȳ)
         return _make_j′vp_call(config.fdm, call, ȳ, primals, is_ignored)
--- a/src/rule_config.jl
+++ b/src/rule_config.jl
@@ -16,8 +16,8 @@ function ChainRulesCore.rrule_via_ad(config::ADviaRuleConfig, f, args...; kws...
 end
 
 # For testing this config uses finite differences to evaluate the frule and rrule
-Base.@kwdef struct ADviaFDConfig <: RuleConfig{Union{HasReverseMode, HasForwardsMode}}
-    fdm = FiniteDifferences.central_fdm(5, 1)
+struct ADviaFDConfig <: RuleConfig{Union{HasReverseMode, HasForwardsMode}}
+    fdm
 end
 
 function ChainRulesCore.frule_via_ad(config::ADviaFDConfig, ȧrgs, f, args...; kws...)
--- a/src/rule_config.jl
+++ b/src/rule_config.jl
@@ -14,3 +14,38 @@ function ChainRulesCore.rrule_via_ad(config::ADviaRuleConfig, f, args...; kws...
     ret === nothing && throw(MethodError(rrule, (f, args...)))
     return ret
 end
+
+# For testing this config uses finite differences to evaluate the frule and rrule
+struct ADviaFDConfig <: RuleConfig{Union{HasReverseMode, HasForwardsMode}}
+    fdm # TODO: add a default
+end
+
+function ChainRulesCore.frule_via_ad(config::ADviaFDConfig, ȧrgs, f, args...; kws...)
+
+    call_on_copy(f, xs...) = deepcopy(f)(deepcopy(xs)...; deepcopy(kws)...)
+
+    primals_and_tangents = auto_primal_and_tangent.((f, args...))
+    primals = primal.(primals_and_tangents)
+    is_ignored = isa.(ȧrgs, NoTangent)
+
+    Ω = call_on_copy(f, args...)
+    ΔΩ = _make_jvp_call(config.fdm, call_on_copy, Ω, primals, ȧrgs, is_ignored)
+
+    return Ω, ΔΩ
+end
+
+function ChainRulesCore.rrule_via_ad(config::ADviaFDConfig, f, args...; kws...)
+
+    call(f, xs...) = f(xs...; kws...)
+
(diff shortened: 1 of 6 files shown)


src/rule_config.jl
```diff
@@ -14,3 +14,38 @@ function ChainRulesCore.rrule_via_ad(config::ADviaRuleConfig, f, args...; kws...
     ret === nothing && throw(MethodError(rrule, (f, args...)))
     return ret
 end
+
+# For testing this config uses finite differences to evaluate the frule and rrule
+Base.@kwdef struct ADviaFDConfig <: RuleConfig{Union{HasReverseMode, HasForwardsMode}}
```
oxinabox, line 19: why the `@kwdef` ?

src/rule_config.jl
```diff
@@ -14,3 +14,38 @@ function ChainRulesCore.rrule_via_ad(config::ADviaRuleConfig, f, args...; kws...
...
+    fdm = FiniteDifferences.central_fdm(5, 1)
+end
+
+function ChainRulesCore.frule_via_ad(config::ADviaFDConfig, ȧrgs, f, args...; kws...)
+
+    call_on_copy(f, xs...) = deepcopy(f)(deepcopy(xs)...; deepcopy(kws)...)
+
+    primals_and_tangents = auto_primal_and_tangent.((f, args...))
+    primals = primal.(primals_and_tangents)
```
oxinabox, line 28: Why is this not
```suggestion
    primals = (f, args...)
```
?

src/rule_config.jl
```diff
@@ -14,3 +14,38 @@ function ChainRulesCore.rrule_via_ad(config::ADviaRuleConfig, f, args...; kws...
...
+end
+
+function ChainRulesCore.rrule_via_ad(config::ADviaFDConfig, f, args...; kws...)
+
+    call(f, xs...) = f(xs...; kws...)
+
+    primals_and_tangents = auto_primal_and_tangent.((f, args...))
+    primals = primal.(primals_and_tangents)
+    accum_cotangents = tangent.(primals_and_tangents)
+    is_ignored = isa.(accum_cotangents, NoTangent)
```
oxinabox, line 44: Do we need to fill in `is_ignored`?
Or would it be ok to return invalid derivatives for these ones.
I suspect the answer is that it would not be OK.
But I wish it wasn't.

I think this section should have a comment explaining that the reason we are calling `auto_primal_and_tangent` is just to try and workout which things we should set `is_ignored` for.

oxinabox (commented):
should this try and use a `rrule`/`frule` first then only fallback to using the FiniteDifferences if that returns `nothing`?

src/rule_config.jl
mzgubic: I think we do need it, because otherwise we get dimension mismatch in `_make_j′vp_call`

Added a comment that explains the block

mzgubic:
>should this try and use a rrule/frule first then only fallback to using the FiniteDifferences if that returns nothing?

Do we ever want to avoid using the rule and make sure to use FD for the inner rule? I don't think so.

I think this means we can just combine the `ADviaRuleConfig` and `ADviaFDConfig` into a `TestConfig`, which does what you describe? And it's not a breaking change because it's not exported?

oxinabox:
> I think this means we can just combine the ADviaRuleConfig and ADviaFDConfig into a TestConfig, which does what you describe?

Yeah, I guess so.

> And it's not a breaking change because it's not exported?

It is, because people use it directly. Don't they?
But we can `@deprecate_binding ADviaRuleConfig TestConfig false`

src/rule_config.jl
```diff
@@ -14,3 +14,38 @@ function ChainRulesCore.rrule_via_ad(config::ADviaRuleConfig, f, args...; kws...
...
+
+function ChainRulesCore.rrule_via_ad(config::ADviaFDConfig, f, args...; kws...)
+
+    call(f, xs...) = f(xs...; kws...)
+
+    # this block is here just to work out which tangents should be ignored
+    primals_and_tangents = auto_primal_and_tangent.((f, args...))
+    primals = primal.(primals_and_tangents)
+    accum_cotangents = tangent.(primals_and_tangents)
+    is_ignored = isa.(accum_cotangents, NoTangent)
```
oxinabox, line 44: Shall we avoid creating a variable that is never used again?
The following fuses
```suggestion
    is_ignored = isa.(tangent.(primals_and_tangents), NoTangent)
```

oxinabox (approved):
merge when happy

src/testers.jl
```diff
@@ -92,7 +92,7 @@ end
    - All remaining keyword arguments are passed to `isapprox`.
 """
 function test_frule(args...; kwargs...)
-    config = ChainRulesTestUtils.ADviaRuleConfig()
+    config = ChainRulesTestUtils.TestConfig()
     test_frule(config, args...; kwargs...)
```
oxinabox, line 96:
```suggestion
    test_frule(TestConfig(), args...; kwargs...)
```

test/testers.jl
```diff
@@ -285,11 +285,11 @@ end
 
     @testset "single input, multiple output" begin
         simo(x) = (x, 2x)
-        function ChainRulesCore.rrule(simo, x)
+        function ChainRulesCore.rrule(::typeof(simo), x)
```
oxinabox, line 288: lol

oxinabox (approved):

test/rule_config.jl
```diff
@@ -49,4 +49,66 @@ struct MySpecialConfig <: RuleConfig{Union{MySpecialTrait}} end
...
+            inner, inner_dot = frule_via_ad(config, (ḟ, ẋ, NoTangent()), f, x, 3.2)
+            return 2 * inner, 2 * inner_dot
+        end
+
+        config = ChainRulesTestUtils.TestConfig()
+        test_rrule(config, outer, inner, rand(); rrule_f=rrule_via_ad, check_inferred=false)
+        test_frule(config, outer, inner, rand(); frule_f=frule_via_ad, check_inferred=false)
+    end
+
+    @testset "Catch incorrect rules" begin
```
mzgubic, line 106: @oxinabox FYI there was actually a gross bug in the original implementation.

This `TestConfig` will be used to test rules for higher order functions `f(inner, array)` where we don't care that much for the rule for `inner` and want to do finite differencing for it out of convenience.

However, if we _don't check whether a rule exists_ inside `f/rrule_via_ad` using the `TestConfig`, we are actually ignoring the outer rule implementation (for `f`) entirely, and always pass the tests because finite differencing is done on the outer rule 😅 

I will wait for you to reply before merging since this was a significant development (even if not a significant change to the code)

oxinabox: oh

PR #247, opened 2022-05-23
Pass keyword arguments to `test_approx` when checking chunks

devmotion:
Just ran into a test issue where I need `nans=true` as keyword argument to `isapprox` (cf https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/220) but it was not forwarded when testing pullbacks with chunks.

### DIFF
--- a/src/testers.jl
+++ b/src/testers.jl
@@ -227,7 +227,7 @@ function test_rrule(
         end
 
         if check_thunked_output_tangent
-            test_approx(ad_cotangents, pullback(@thunk(ȳ)), "pulling back a thunk:")
+            test_approx(ad_cotangents, pullback(@thunk(ȳ)), "pulling back a thunk:"; isapprox_kwargs...)
             check_inferred && _test_inferred(pullback, @thunk(ȳ))
         end
     end  # top-level testset


oxinabox (approved):
This is definitely the correct behavior, regardless of if we do #220 or not.

I am not sure what is with the CI errors in Julia 1.0 on some plateforms but not others though.

devmotion:
> I am not sure what is with the CI errors in Julia 1.0 on some plateforms but not others though.

These test errors all seem to happen in checks that involve thunks. So I assume these tests were a bit brittle before but passed (usually) with the default `atol=0` and `rtol=sqrt(eps())` settings. With this PR, however, they are also run with `atol = rtol = 1e-9` which probably uncovers these issues.

devmotion:
A second observation is that the failing tests involve comparisons of `Float64` and `Float32`. This might explain why it's an issue on some architectures only, and why `atol = rtol = 1e-9` might be problematic.

devmotion:
Test errors should be fixed now. If there are no additional comments, I'll merge when the CR downstream tests pass.

Issue #229, opened 2022-01-15
`test_approx` has ambiguities


mcabbott:
```julia
julia> ChainRulesTestUtils.test_approx(Tangent{Int}(; x=1), ZeroTangent())
ERROR: MethodError: test_approx(::Tangent{Int64, NamedTuple{(:x,), Tuple{Int64}}}, ::ZeroTangent) is ambiguous. Candidates:
  test_approx(x, ::AbstractZero) in ChainRulesTestUtils at /Users/me/.julia/packages/ChainRulesTestUtils/8dFTY/src/check_result.jl:34
  test_approx(actual::Tangent{P, T}, expected) where {T, P} in ChainRulesTestUtils at /Users/me/.julia/packages/ChainRulesTestUtils/8dFTY/src/check_result.jl:108
Possible fix, define
  test_approx(::Tangent{P, T}, ::AbstractZero) where {T, P}
Stacktrace:
```
That's a made-up example, the real one gave:
```
Error During Test at /Users/me/.julia/packages/ChainRulesTestUtils/8dFTY/src/testers.jl:195
  Got exception outside of a @test
  MethodError: (::ChainRulesTestUtils.var"#test_approx##kw")(::NamedTuple{(:rtol, :atol), Tuple{Float64, Float64}}, ::typeof(test_approx), ::ChainRulesCore.NoTangent, ::ChainRulesCore.Tangent{Tuple{Float64, Float64, Float64}, Tuple{Float64, Float64, Float64}}) is ambiguous. Candidates:
    (var"#s103"::ChainRulesTestUtils.var"#test_approx##kw")(::Any, ::typeof(test_approx), x, y::ChainRulesCore.Tangent) in ChainRulesTestUtils at /Users/me/.julia/packages/ChainRulesTestUtils/8dFTY/src/check_result.jl:124
    (var"#s102"::ChainRulesTestUtils.var"#test_approx##kw")(::Any, ::typeof(test_approx), ::ChainRulesCore.AbstractZero, x) in ChainRulesTestUtils at /Users/me/.julia/packages/ChainRulesTestUtils/8dFTY/src/check_result.jl:33
  Possible fix, define
    (::ChainRulesTestUtils.var"#test_approx##kw")(::Any, ::typeof(test_approx), ::ChainRulesCore.AbstractZero, ::ChainRulesCore.Tangent)
```

PR #253, opened 2022-07-08
Resolve `test_approx` ambiguities

### LINKED ISSUE #252: `test_approx` ambiguities
Found in the wild here: https://github.com/JuliaDiff/ChainRules.jl/pull/642#discussion_r915406061

And some similar test cases that fail:
```julia
test_approx(
    @thunk(Tangent{Tuple{Float64,Float64}}(1.0, 2.0)),
    Tangent{Tuple{Float64,Float64}}(1.0, 2.0),
)

test_approx(@thunk(ZeroTangent()), ZeroTangent())

test_approx(ZeroTangent(), @thunk(ZeroTangent()))

test_approx(
    Tangent{Tuple{Float64,Float64}}(ZeroTangent(), NoTangent()),
    NoTangent(),
)

test_approx(
    NoTangent(),
    Tangent{Tuple{Float64,Float64}}(ZeroTangent(), NoTangent()),
)
```

mzgubic:
Closes #252

### DIFF
--- a/test/check_result.jl
+++ b/test/check_result.jl
@@ -83,6 +83,20 @@ end
                 Tangent{Tuple{Float64,Float64}}(1.0, 2.0),
                 @thunk(Tangent{Tuple{Float64,Float64}}(1.0, 2.0)),
             )
+            test_approx(
+                @thunk(Tangent{Tuple{Float64,Float64}}(1.0, 2.0)),
+                Tangent{Tuple{Float64,Float64}}(1.0, 2.0),
+            )
+            test_approx(@thunk(ZeroTangent()), ZeroTangent())
+            test_approx(ZeroTangent(), @thunk(ZeroTangent()))
+            test_approx(
+                Tangent{Tuple{Float64,Float64}}(ZeroTangent(), NoTangent()),
+                NoTangent(),
+            )
+            test_approx(
+                NoTangent(),
+                Tangent{Tuple{Float64,Float64}}(ZeroTangent(), NoTangent()),
+            )
         end
         @testset "negative case" begin
             @test fails(() -> test_approx(1.0, 2.0))
--- a/test/check_result.jl
+++ b/test/check_result.jl
@@ -90,12 +90,12 @@ end
             test_approx(@thunk(ZeroTangent()), ZeroTangent())
             test_approx(ZeroTangent(), @thunk(ZeroTangent()))
             test_approx(
-                Tangent{Tuple{Float64,Float64}}(ZeroTangent(), NoTangent()),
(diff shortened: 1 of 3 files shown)


test/check_result.jl
```diff
@@ -83,6 +83,20 @@ end
                 Tangent{Tuple{Float64,Float64}}(1.0, 2.0),
                 @thunk(Tangent{Tuple{Float64,Float64}}(1.0, 2.0)),
             )
+            test_approx(
+                @thunk(Tangent{Tuple{Float64,Float64}}(1.0, 2.0)),
+                Tangent{Tuple{Float64,Float64}}(1.0, 2.0),
+            )
+            test_approx(@thunk(ZeroTangent()), ZeroTangent())
+            test_approx(ZeroTangent(), @thunk(ZeroTangent()))
+            test_approx(
```
mzgubic, line 92: Do we want this? I think we do?

test/check_result.jl
```diff
@@ -83,6 +83,20 @@ end
...
+                Tangent{Tuple{Float64,Float64}}(ZeroTangent(), NoTangent()),
+                NoTangent(),
+            )
+            test_approx(
+                NoTangent(),
+                Tangent{Tuple{Float64,Float64}}(ZeroTangent(), NoTangent()),
+            )
```
mzgubic, line 99: Do we want this? I think we do

oxinabox (approved):

mcabbott:
Fixes the example from #229 too:
```
julia> ChainRulesTestUtils.test_approx(Tangent{Int}(; x=1), ZeroTangent())
false

(jl_rp9LKs) pkg> st
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_rp9LKs/Project.toml`
  [cdddcdb0] ChainRulesTestUtils v1.8.2 `https://github.com/JuliaDiff/ChainRulesTestUtils.jl#mz/testapprox`
```

test/check_result.jl
mzgubic: In practice I think what matters is that `NoTangent()` is the same as `Tangent` of just `NoTangent()`s.

E.g. the choice similar to the one we made in
```julia
julia> rand_tangent((1, 2))
NoTangent()
```

I added on the `ZeroTangent` on purely practical grounds, since they are effectively the same. But I can't justify it either, so we can push that decision down the road when we deal with automating testing a bit better, and restrict to `NoTangent` for now.

Does that sound good to merge?

oxinabox: That sounds good

PR #273, opened 2023-03-08
Allow editing the testset name

theogf:
Solves #272 

Add a keyword argument "`testset_name`" to `test_scalar`, `test_frule` and `test_rrule` to allow replacing the default name.
It still stays an opt-in option and is mostly useful when `f` is anonymous and its naming is too large to be displayed correctly.

PR #274, opened 2023-03-08
Do not use `isnothing` for Julia 1.0 compat

devmotion:
#273 broke downstream tests, e.g., in AbstractFFTs, since it broke compatibility with Julia 1.0 (and maybe some other versions as well, I don't remember when `isnothing` was added).

This PR should fix it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRulesTestUtils"
 uuid = "cdddcdb0-9152-4a09-a978-84456f9df70a"
-version = "1.10.0"
+version = "1.10.1"
 
 [deps]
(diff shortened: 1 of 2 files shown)


devmotion:
It seems it was not caught by CI since the Github action is disabled?

gaurav-arya:
Julia 1.0 should also be added to appropriate tests too, right? (BlockDiagonals.jl seems to support 1.0 and already has an integration test, so perhaps that integration test should run on 1.0 too?)

devmotion:
I re-enabled it and triggered it on this PR.

gaurav-arya:
Okay, you are again ahead of me

oxinabox:
Good catch.
I forgot we supported Julia 1.0

PR #287, opened 2023-09-28
Monkey patch Test.scrub_backtrace

oxinabox:
Problem is 
the `@test` macro is that it scrubs backtraces to remove stuff from deeper down the stack than the location of the macro, and since `test_rrule` etc use those macros directly the back trace no longer includes the call site in the users tests.

Simple solution is to just stop scrubbing backtraces entirely.
If people like this we could follow up with something more sophisticated.
But lets have that as a seperate PR
 
This does mean whenever tests are run this warning will show.

```
     Testing Running tests...
WARNING: Method definition scrub_backtrace(Any, Any, Any) in module Test at /home/oxinabox/dist/julia-cedar/vanilla/latest/share/julia/stdlib/v1.11/Test/src/Test.jl:89 overwritten at /home/oxinabox/JuliaEnvs/ChainRulesTestUtils.jl/src/ChainRulesTestUtils.jl:26.
Testing ChainRules.jl
```
Could surpress that though.

### DIFF
--- a/src/ChainRulesTestUtils.jl
+++ b/src/ChainRulesTestUtils.jl
@@ -23,6 +23,8 @@ function __init__()
     # Try to disable backtrace scrubbing so that full failures are shown
     try
         isdefined(Test, :scrub_backtrace) || error("Test.scrub_backtrace not defined")
+        # depending on julia version or one or the other of these will be hit
+        @eval Test scrub_backtrace(bt,) = bt  # make it do nothing
         @eval Test scrub_backtrace(bt, file_ts, file_t) = bt  # make it do nothing
     catch err
         @warn "Failed to monkey=patch scrub_backtrace. Code is functional but stacktraces may be less useful" exception=(err, catch_backtrace())
--- a/src/ChainRulesTestUtils.jl
+++ b/src/ChainRulesTestUtils.jl
@@ -10,6 +10,7 @@ using FiniteDifferences: to_vec
 using LinearAlgebra
 using Random
 using Test
+using Suppressor
 
 export TestIterator
 export test_approx, test_scalar, test_frule, test_rrule, generate_well_conditioned_matrix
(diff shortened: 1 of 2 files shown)


src/ChainRulesTestUtils.jl
```diff
@@ -17,7 +17,17 @@ export ⊢, rand_tangent
...
 export test_method_tables
 
-__init__() = init_test_inferred_setting!()
+function __init__()
+    init_test_inferred_setting!()
+
+    # Try to disable backtrace scrubbing so that full failures are shown
+    try
+        isdefined(Test, :scrub_backtrace) || error("Test.scrub_backtrace not defined")
+        @eval Test scrub_backtrace(bt, file_ts, file_t) = bt  # make it do nothing
```
oxinabox, line 26:
```suggestion
        # depending on julia version or one or the other of these will be hit
        @eval Test scrub_backtrace(bt,) = bt  # make it do nothing
        @eval Test scrub_backtrace(bt, file_ts, file_t) = bt  # make it do nothing
```

oxinabox:
I tried adding a `with_logger(NullLogger()` around the `@eval` and it didn't silence the warning.
:-(

willtebbutt (approved):
LGTM. Tested locally on some of my own code, and seems to work fine. Play nicely with Revise etc (as you would expect), so seems to be fine.

src/ChainRulesTestUtils.jl
```diff
@@ -10,14 +10,29 @@ using FiniteDifferences: to_vec
...
+    # Try to disable backtrace scrubbing so that full failures are shown
+    try
+        isdefined(Test, :scrub_backtrace) || error("Test.scrub_backtrace not defined")
+        @suppress begin  # mute warning about monkey-patching
+            # depending on julia version or one or the other of these will be hit
+            @eval Test scrub_backtrace(bt,) = bt  # make it do nothing
+            @eval Test scrub_backtrace(bt, file_ts, file_t) = bt  # make it do nothing
+        end
+    catch err
+        @warn "Failed to monkey=patch scrub_backtrace. Code is functional but stacktraces may be less useful" exception=(err, catch_backtrace())
```
oxinabox, line 33:
```suggestion
        @warn "Failed to monkey-patch scrub_backtrace. Code is functional but stacktraces may be less useful" exception=(err, catch_backtrace())
```

PR #294, opened 2025-07-03
Fix imports of Test functions on versions >1.12

lkdvos:
After https://github.com/JuliaLang/julia/pull/57839, the `Test.eval_test` function no longer exists and is split into `eval_test_comparison` and `eval_test_function`. This fixes the imports, such that CI tests on Julia nightly can run.

It might be better to not depend on internals in the long run, but I don't currently have the time to fix this.


vtjnash:
LGTM

andreasnoack:
Fixed in https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/293

PR #295, opened 2026-03-29
Add compat with ChainRulesCore v2

JamesWrigley:
Tested manually by `dev`ing v2 and https://github.com/JuliaDiff/FiniteDifferences.jl/pull/248.

### DIFF
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


JamesWrigley:
I did the same integration test fix as in https://github.com/JuliaDiff/FiniteDifferences.jl/pull/248, which should fix the SpecialFunctions integration test. The BlockDiagonals test should be fixed by https://github.com/JuliaArrays/BlockDiagonals.jl/pull/151.

